PR #12, opened 2020-11-19
Docker

dgerblick:
Added python script to install dependencies ros. Works by searching for dependency tags in package.xml files.

Added arguments to a few docker launch files to pass to the command.

Also general tools from apt


jrbaxter0 (approved):
After a long discussion with Daniel, this seems like a good move in order to clean stuff up and make our development process a little easier.

PR #15, opened 2021-02-16
clean up scripts

jrbaxter0:


dgerblick (commented):
Check the changed variable name in docker-join.sh. Also, what's going on with the changes to docker-run.sh?

jrbaxter0:
> Check the changed variable name in docker-join.sh.
Fixed that now

> Also, what's going on with the changes to docker-run.sh?
That is to allow the cleaner method of passing in parameters. Can do `--gpus all`, `--net host`, '--privileged` all without needing weird redundancy in the script. Will be more flexible and intuitive IMO

jrbaxter0:
This has been pulled into #16 since it is heavily reliant on it.

PR #16, opened 2021-02-17
Ai docker

jrbaxter0:
This updates the Docker image so that it can be used for high level work.

Primary additions:
- python 3 support with ROS
- Tensorflow (with CUDA GPU support)
- Pytorch (with CUDA GPU support)

Secondary additions:
- tmux and a few other tools
- error handling in build scripts
- arbitrary docker run options (such as `--gpus all`)


MeenakshiMcNamara (approved):
I was able to get pytorch to run using the gpu of the super computer, so it seems good to me!

dgerblick (changes requested):
--privileged is removed from docker-run, which NEEDS to be fixed to access devices (like cameras).
--hostname is also removed, which is fine I guess, but seems weird to me.

Installing of project software dependencies (Dockerfile, lines 50-51) seems to me like a less optimal solution to what was previously there. The usb_cam installation is redundant in either case, since it comes preinstalled with ros-melodic-desktop-full (though I guess it's still nice to have examples)

jrbaxter0:
@dgerblick Regarding `--privileged`, that was intentional. Due to the new `$@` syntax, people can pass that in on their own. Since most people running this will not need it to access cameras, it seemed like a good idea not to open up that potential vulnerability. The main concern is people running things as root in the container and modifying important system files by mistake due to them being accessible to the container.

Regarding, `--hostname`, that was also intentional. When you do `--net host`, it messes things up to have a different hostname than the bare-metal computer. This will resolve that, without any significant drawbacks. I only did `--hostname` originally to make things look nicer in bash. But we use zsh and you don't even see it.

Regarding `rosdep`, I had issues getting it working. It seems it doesn't come standard with ROS installs. When I did install it, it had trouble resolving to packages. I'm not very concerned about it though, because our use case doesn't really require it. The package names in APT are not changing, and we only support one operating system. We also aren't building a lot of 3rd party code from source.

dgerblick (approved):
someone should look over the changes I made, make sure everything still works

dgerblick:
> @dgerblick Regarding `--privileged`, that was intentional. Due to the new `$@` syntax, people can pass that in on their own. Since most people running this will not need it to access cameras, it seemed like a good idea not to open up that potential vulnerability. The main concern is people running things as root in the container and modifying important system files by mistake due to them being accessible to the container.
> 
> Regarding, `--hostname`, that was also intentional. When you do `--net host`, it messes things up to have a different hostname than the bare-metal computer. This will resolve that, without any significant drawbacks. I only did `--hostname` originally to make things look nicer in bash. But we use zsh and you don't even see it.
> 
> Regarding `rosdep`, I had issues getting it working. It seems it doesn't come standard with ROS installs. When I did install it, it had trouble resolving to packages. I'm not very concerned about it though, because our use case doesn't really require it. The package names in APT are not changing, and we only support one operating system. We also aren't building a lot of 3rd party code from source.

Sounds good, makes sense to me

jrbaxter0:
@dgerblick These lines allow us to greatly simplify the edits you had:
```
chsh -s /bin/zsh $USER
CMD /bin/zsh
```

Sets the default shell for the user to zsh (I did this because I really like tmux, and this makes the tmux shells zsh instead of bash), and make the default command to run zsh if nothing is given (by the environment variable being unset).

This is changed in the newest Dockerfile, and doesn't work if you're using the old one with the new scripts.

jrbaxter0:
Everything is tested to work, when running on the supercomputer. I haven't built it locally since it takes a lot longer, but I am confident in it.

PR #17, opened 2021-02-18
fixed orientation of path points because they were pointing in the op…

MeenakshiMcNamara:
…posite direction they should be


hmccarty (approved):
LGTM!

adte01234 (approved):

PR #19, opened 2021-02-27
Midlevel meenakshi

MeenakshiMcNamara:
Can you guys check this out? I know the goal that the high level planner is using is wrong, but I think I have the interface working properly (also, if you could replace the yaml file goal with the correct one that would be great!)


adte01234:
image

Are the way points supposed to look like this? Also, the car isn't moving. I forget if we solved this last time.
Apart from that this looks fine.

adte01234 (approved):

hmccarty (changes requested):
I tried setting the goal point to what was in the sim however it seemed to throw a fit about types. We should ensure that it works when we pass in a float for the goal point.

PR #32, opened 2021-03-30
Ball Tracking

ajernandes:
Implemented the transformation onto the whole field coordinate plane; should be ready to test on Saturday.


dgerblick:
I'm still not sold on the method for localization here. I think doing a simple offset would introduce too much distortion near the edges of the frame. But that is what this weekend is for, and I'd love to be proven wrong.

Aside from that, there are some things in here regarding coding style that seem... weird, but nothing that would impact functionality. Unless James has anything to add, I think we're good to merge.

ajernandes:
> I'm still not sold on the method for localization here. I think doing a simple offset would introduce too much distortion near the edges of the frame. But that is what this weekend is for, and I'd love to be proven wrong.
We are actually not doing that, in `BallDetection.cpp` on line 113, there is this bit of code: `cv::Point3d cam = camera.projectPixelTo3dRay(cv::Point2d(centerX, centerY));`.
The `projectPixelTo3dRay` method ([documentation ](https://docs.ros.org/en/api/image_geometry/html/c++/classimage__geometry_1_1PinholeCameraModel.html#a40ad5d3b964d63f41232428fb96376fe)) Is what I'm using, along with some trig. For what it's worth, I've tested this at home with a tape measure and it was basically spot-on; we will be able to check this further this weekend. I know the coding is a bit weird, particularly the putting the size on the z and some of the weird stuff I did. Once this works, I'll work on straightening things out.

dgerblick:
Oh I'm dumb. I saw lines 131/132 and vaguely linked it to something we talked about a while ago. I completely missed the projectPixelTo3dRay line when I was skimming the code. Sorry about that.

ajernandes:
Ah, I see.

Those lines are for taking the coordinates centered at the point directly below the camera and translating those to the origin or the entire field.

PR #40, opened 2021-04-17
Delay Reduction

dgerblick:
Let's see how well this works

- Not tested with multiple cameras.
- Added blocking calls for (some) tf queries,
- Made AprilTag Detection node into nodelet


jrbaxter0:
@dgerblick, I'm trying to clean up some branches. What is the state of this one? Is this something you're still working on, or is it a failed experiment?

PR #45, opened 2021-08-25
Added adrian's notes

hmccarty:
Adrian originally added this doc to the wiki, but given its level of detail, it's better suited here.

This is connected to the push to make the wiki more 'functionally detailed' versus 'technically detailed'.


jrbaxter0:
documentation is good :)

jrbaxter0 (approved):
oops, my last one was a comment, not an approval

PR #55, opened 2021-09-26
Gym api

jrbaxter0:
Works well enough but needs a little work still. Merge into main to create a clean starting point.


jrbaxter0:
The merge was funny, so I undid it. I'll try redoing it and see where things went wrong

jrbaxter0:
I just did these merges manually

PR #57, opened 2021-09-27
Autonomy main

jrbaxter0:
This is becoming somewhat stable, so we should merge it in before things diverge too much.


jrbaxter0:
@dgerblick Can you approve the Dockerfile changes and see if we can get that rebuilt as part of the fixes we're doing there. It may also be nice to add `htop` and put those project dependency installs up top too. And since we aren't using Tensorflow, we may not even need to build off of the CUDA image. I really don't know.

jrbaxter0:
@halcdev Can you see if it renders on Mac? It renders on Ubuntu w/ GPU pass through (If I downgrade Pyglet to a non-broken version), and it surprisingly renders on WSL2.

You will need to run:
```
sudo apt install python3-tk
pip3 install stable-baselines3
```
to make up for the changes in the Dockerfile.

jrbaxter0:
@halcdev specific commands to test are:
`roslaunch rocket_league_drl cartpole.launch`
`rosrun rocket_league_drl train.py`

halcdev:
> @halcdev Can you see if it renders on Mac? It renders on Ubuntu w/ GPU pass through (If I downgrade Pyglet to a non-broken version), and it surprisingly renders on WSL2.
> 
> You will need to run:
> 
> ```
> sudo apt install python3-tk
> pip3 install stable-baselines3
> ```
> 
> to make up for the changes in the Dockerfile.

Stable-Baselines3 worked fine, but entering root password with sudo is not working at all for some reason, so I can't get the command to run with the correct permissions.

jrbaxter0:
> 
> 
> > @halcdev Can you see if it renders on Mac? It renders on Ubuntu w/ GPU pass through (If I downgrade Pyglet to a non-broken version), and it surprisingly renders on WSL2.
> > You will need to run:
> > ```
> > sudo apt install python3-tk
> > pip3 install stable-baselines3
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to make up for the changes in the Dockerfile.
> 
> Stable-Baselines3 worked fine, but entering root password with sudo is not working at all for some reason, so I can't get the command to run with the correct permissions.

Password is `robot`. Try killing and restarting the container?

jrbaxter0:
@dgerblick I think this may try to install torch early since it is required by stable-baselines3 iirc

purduearc:
re-opening so the action runs

jrbaxter0:
This should work now

PR #58, opened 2021-09-29
fixed noetic upgrade issues

hmccarty:
was working locally after these changes. might be more to change, haven't investigated.


jrbaxter0 (approved):
I don't track any of these files, so any changes you want to make are fine by me. Do you have anything specific you wanted me to look at?

PR #63, opened 2021-10-20
Add CUDA support to arm images

dgerblick:
Still a WIP. Don't merge this yet I just wanted to trigger the autobuild for later testing.


jrbaxter0:
@dgerblick I'm just going to close this since we aren't going to use ARM for the forseeable future. We'll have to test at some point to see if R2D2 can run our network even without a GPU. It may be easier to buy an nVidia GPU if we don't like what we see...

PR #65, opened 2021-10-31
plotter for autonomy data

jrbaxter0:
Integrate into the existing plotter code.


halcdev (approved):
Data is plotted correctly in catkin_ws/data/snakesim!

PR #68, opened 2021-11-08
Autonomy cleanup

jrbaxter0:
Fixes the issue with plotting


halcdev:
New plotting works. If a trained model is re-loaded from a .zip file, it overrides the plot.

halcdev (approved):
New plotting works successfully after testing.

PR #69, opened 2021-11-14
TF breakup + Magic

dgerblick:
- Added smoothing to the camera's position (with magic).
- Still outputs transforms in 3d though, next goal is to transform into 2d.
- Seems to be slightly smoother between cameras
- Added Eigen as a dependency for Linear Algebra magic (PR is mainly to trigger the autobuild for the integration test)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -51,6 +51,16 @@ RUN pip3 install pybullet stable-baselines3 && \
     ros-${ROS_DISTRO}-pointgrey-camera-driver && \
     rm -rf /var/lib/apt/lists/* /root/.cache/pip/
 
+# Install eigen
+RUN mkdir eigen && \
+    cd eigen && \
+    wget https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.tar.gz && \
+    tar xzf eigen-3.4.0.tar.gz && \
+    cmake eigen-3.4.0 && \
+    make install && \
+    cd .. && \
+    rm -rf eigen
+
 # create user
 ARG USER=arc
 ARG PW=password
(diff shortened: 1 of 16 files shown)


jrbaxter0:
@dgerblick @ajernandes Can you add the changes from today, then I'll approve?

jrbaxter0:
Oh, I see its already there

docker/Dockerfile
```diff
@@ -51,6 +51,16 @@ RUN pip3 install pybullet stable-baselines3 && \
     ros-${ROS_DISTRO}-pointgrey-camera-driver && \
     rm -rf /var/lib/apt/lists/* /root/.cache/pip/
 
+# Install eigen
+RUN mkdir eigen && \
```
jrbaxter0, line 55: @dgerblick can't we just get this through APT? seems like overkill

docker/Dockerfile.local
```diff
@@ -1,4 +1,5 @@
-FROM purduearc/rocket-league
+ARG TAG=local
+FROM purduearc/rocket-league:${TAG}
```
jrbaxter0, line 2: Is this meant to be here, or just for local testing?

jrbaxter0 (changes requested):
Looks good, but address two questions pls

docker/Dockerfile
dgerblick: Didn't realize apt had eigen, oops

docker/Dockerfile.local
dgerblick: Typo, line 1 is supposed to be latest. So you can run ./docker-run --build-arg TAG=69 and the docker image for this PR

jrbaxter0 (approved):

PR #70, opened 2021-11-14
Sim polishing

hmccarty:
- Randomized positions of ball and car
- Visualizer able to display walls & rotation


jrbaxter0:
I think something is funny with how the car moves. We will need to look into that further, but the goals of this PR are met.

jrbaxter0 (approved):

PR #72, opened 2021-11-21
Merge ball tracking into main

ajernandes:
Tested today, worked. Considering switching ball color due to some noise from jeans & USB cables.


jrbaxter0 (approved):
Lgtm. We'll need to update the Odom node in the future as discussed

PR #73, opened 2021-12-01
Sim polishing

hmccarty:
- Removed bug in visualizer
- Added framework for testing sim (still needs some work)

Mainly pulling in to resolve visualizer for training.


jrbaxter0:
Seems to function

jrbaxter0 (approved):

PR #75, opened 2022-01-14
Conventions redo

jrbaxter0:
This is a big renaming / reorganization of many project files. This is useful because the project has grown such that each package is completely different from all the others and there is a mess of dependencies and launch files strewn everywhere.

Some of the key things it does:
- merge some similar packages together
- rename packages with a consistent scheme
- enforce conventions within packages

Future things:
- enforce standardized readme
- continue cleaning up launch files
- remove heavily depreciated code

Work that must be done before merging:
- ensure there are no large breaking changes (small ones are OK, we can fix them as we go)
- ensure the merging / renaming / standardization is OK with the code owners

Timeline
- this should be merged before we undertake any new big changes this semester (ideally in the next few days)


jrbaxter0:
Everything builds properly and the simulator, visualizer, and trainer all run:
```
catkin clean
catkin build
roslaunch rktl_autonomy rocket_league.launch
```

jrbaxter0:
I think the `rktl_perception` package needs to be looked at. There is just a lot of C code and it is hard to tell what is responsible for what, and which parts are even still in use.

dgerblick:
> I think the `rktl_perception` package needs to be looked at. There is just a lot of C code and it is hard to tell what is responsible for what, and which parts are even still in use.

I just removed the unused files, code still needs refactor but that's a different conversation

ajernandes (approved):

halcdev (approved):
💪 power

PR #78, opened 2022-01-20
Autonomy rewards

jrbaxter0:
This adds in two new rewards, a wall penalty and a reverse penalty. These can be used to discourage behaviors we don't like. Additionally, it adds in test cases (that work well enough, but could use a little more tweaking), and **resolves 2 major bugs in the rewards system that I found through creating the test cases**. That is probably why it trained poorly. We were rewarding it based off of yaw, not location . . .

To run the tests
1. Checkout this branch
2. `catkin build` everything
3. run `catkin test` to test this and the `rktl_test` branch

You may need to run it a second time for the tests to pass. There is some weird behavior with retrying the clocks, which isn't a big issue in practice, but screws up the test cases occasionally since they all need to match. I will fix it eventually.

EDIT:
Another way to run the tests (which is handy for debugging) is:
```
roscore
rostest test/test_step.test --text --reuse-master
```
This allows you to use standard tools in another window to inspect ROS messages and gives you the output in plain text


rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
```diff
@@ -151,16 +153,24 @@ def _get_state(self):
         # Determine reward
         reward = self._CONSTANT_REWARD
 
-        ball_dist_sq = np.sum(np.square(ball[1:3] - car[1:3]))
```
jrbaxter0, line 154: This is one of those critical issues I mentioned

rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
```diff
@@ -151,16 +153,24 @@ def _get_state(self):
...
+        ball_dist_sq = np.sum(np.square(ball[0:2] - car[0:2]))
         reward += self._BALL_DISTANCE_REWARD * ball_dist_sq
 
-        goal_dist_sq = np.sum(np.square(ball[1:3] - np.array([self._FIELD_HEIGHT, 0])))
+        goal_dist_sq = np.sum(np.square(ball[0:2] - np.array([self._FIELD_HEIGHT/2, 0])))
```
jrbaxter0, line 159: This is the other `FIELD_HEIGHT/2`

hmccarty (commented):
Looks good, would be interested to see if we can run it on the super comp before merging

hmccarty (approved):

PR #79, opened 2022-01-21
Wrong goal punishment

jrbaxter0:
This adds an optional penalty for when the car scores in the wrong goal and adds a test case to confirm this behavior. It also improves the test case infrastructure so that they don't randomly fail (for this specific test). The tests can be run with `catkin test`


jrbaxter0:
@hmccarty @dgerblick Can we set up a workflow with GitHub so that it runs `catkin build` and `catkin test` inside the Docker container and verifies that they build all packages?

jrbaxter0:
Also, @hmccarty I'm pretty comfortable with how `rostest` works now, so we can take a look at yours together and see if we can get them working

hmccarty (approved):
Looks good, tests run locally.

PR #80, opened 2022-01-21
Create catkin.yml

jrbaxter0:
@dgerblick Any idea why `DOCKER_CMD="/bin/zsh -c 'cd catkin_ws && catkin build'" ./docker/docker-run.sh` does not work? I get `zsh:1: unmatched '`


jrbaxter0:
For reference, running `/bin/zsh -c 'cd catkin_ws && catkin build'` _after_ starting the Docker container in interactive mode works fine

jrbaxter0:
I figured that we could `grep` the output for `[build] Summary: All 10 packages succeeded!`, perhaps replacing `10` for any number using a regex. If it fails, it'll say `[build] Summary: 9 of 10 packages succeeded.`, which wouldn't match the pattern.

.github/workflows/catkin.yml
```diff
@@ -0,0 +1,26 @@
...
+        cd catkin_ws &&
+        catkin build --no-status --force-color &&
+        catkin test --no-status --force-color
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        path: src/rocket_league
+    - name: Build the local Docker image
+      run: ./src/rocket_league/docker/docker-build.sh --build-arg TAG=${{ github.event.number }}
```
jrbaxter0, line 24: We only want to use a non-default tag if the PR forced a new Docker image to be built. And if that was the case, this action needs to wait until the other action finishes.

docker/docker-run.sh
```diff
@@ -5,7 +5,14 @@ REPO_NAME="purduearc/rocket-league"
 CONTAINER_NAME="${CONTAINER_NAME:-arc-rocket-league-dev}"
 echo "mounting host directory $WS_DIR as container directory /home/$USER/catkin_ws"
 
-docker run --rm -it \
+# tty-specific options
+if [ -t 0 -a -t 1 ]
```
jrbaxter0, line 9: Can you explain what this is doing? I assume it adds `-it` if this is an interactive terminal?

dgerblick: Yeah exactly, running docker with -it throws and error if it's not running in a terminal. The action isn't in a terminal, which was causing the error that you ran into

jrbaxter0:
@dgerblick It works now, even though the job failed. That is because I just re-configured it for main and it won't work.

You can see the successful run here: https://github.com/purdue-arc/rocket_league/actions/runs/1734448054
it ran the tests, and the `rktl_autonomy` ones failed, killing the workflow. I have another PR open that fixes those tests.

We may need to modify it very slightly so that it runs dependent jobs if the initial ones are skipped. I'm not 100% sure if we need to explicitly tell it to do so, like `job_d` at this link: https://github.com/actions/runner/issues/491

We will need to test that once merging via a test PR

dgerblick (approved):
Everything looks good to me but I have no clue what I'm talking about. I'm calling it good enough for government work and if there's errors, oh well.

PR #81, opened 2022-01-22
Sim updates

hmccarty:
Fixes a few of the requested items:

- Ball bouncing (lower speeds will larger angles of reflection, but overall it looks much better. more tuning can be done if still see issues)
- Wall sliding (added pseudo-friction to prevent extreme sliding, also looks a lot better but could be further tuned if needed)

Also gave the visualizer a little more flare

Plan on adding updated tests in separate PR


jrbaxter0:
@hmccarty Please include instructions to test in all future PRs. This could be running a test case or manual instructions. I'm going to run the sim and give manual input to the car to see how it behaves.

jrbaxter0 (changes requested):
I like the new field image. Eventually I'd like to update the car too, but that isn't really a pressing issue.

The reason I'm requesting changes is because calling the `sim_reset` service doesn't give the ball a random velocity like when it originally spawns.

I also think the bouncing off walls could use more tweaking, but it is a definite improvement and I don't want to hold back the "ball initial velocity" feature due to that.

hmccarty:
Later today I'm going to refactor the reset and init function to be combined to a single function.

jrbaxter0:
> Later today I'm going to refactor the reset and init function to be combined to a single function.

Do you want me to wait to review until after that? Or do you want to do that as another PR?

hmccarty:
We can go ahead and merge, I'll attach it to the upcoming PR. Honestly was just looking to leave a note for myself, since I'm not at my computer atm.

jrbaxter0:
image
I'm going to merge this in, even though there is a bug in the sim. Both goals result in victory for team A. Took me a very long time to make that happen lol

jrbaxter0:
But the automatic reset seems to work, and the ball bouncing is improved imo

jrbaxter0 (approved):

PR #82, opened 2022-01-24
simple keyboard interface for testing

jrbaxter0:
This is a simple node for using keyboard input to test the sim. Useful since I don't always have an XBOX controller around, and getting to to run on WSL2 isn't worth the effort.

Testing:
```
roslaunch rktl_launch rocket_league_sim.launch
rosrun rktl_sim keyboard_interface
```

All controls are explained in the window.

### DIFF
--- a/rktl_sim/nodes/keyboard_interface
+++ b/rktl_sim/nodes/keyboard_interface
@@ -2,7 +2,7 @@
 """Use keyboard data to control car in sim.
 License:
   BSD 3-Clause License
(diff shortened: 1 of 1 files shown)


hmccarty (commented):
This is cool, works great on my end. Do we want to add this to the sim launch file with an argument?

jrbaxter0:
I don't think so. Curses takes over the entire terminal and I feel like that may mess up other things, like printing errors.

It is also easy enough to launch by itself

hmccarty (approved):

PR #83, opened 2022-01-25
fixed message publication for scoring

hmccarty:
Tested by resetting sim until ball collided with both goals. Top goal published a match_status=3, while bottom published a match_status=4, as expected.


jrbaxter0 (approved):
I'm not going to test it, since it took a long time to make it hit both goals lol.

PR #84, opened 2022-01-25
Replay

jrbaxter0:
This update allows you to replay old weights in real time. This will be needed when we want to eventually run the autonomy system live.

@halcdev The launch file for training has been renamed to `rocket_league_train.launch`

Testing: This needs to be done manually since there won't be weight files on the automated testing system. Simply put `weights.zip` in the `catkin_ws/data/rocketleague` folder, and watch it use them.


halcdev (approved):
`rocket_league_eval.launch` runs successfully when 'weights.zip' is in the `catkin_ws/data/rocketleague` directory. Crashes when it is not.

It opens the vis and only initiates one "run" of the sim.

PR #85, opened 2022-02-01
Vec env

jrbaxter0:
This redesigns the infrastructure for training the agent for the primary goal of allowing multiple environments to run concurrently. This is intended for two main uses cases:
- vectorized environments: A single agent runs several environments concurrently, increasing the rate at which it learns.
- hyperparameter searches: several agents run their own environment(s) concurrently, allowing us to learn which hyperparameters (ex step sizes, reward functions) work best.

The main training script is now entirely separate from ROS. When it creates an environment, that process will launch ROS. If done via the vectorized environments, it will create a new process for it. The env will run its own ROS core on a random free port on the computer. This allows it to run the simulated time at whatever speed is necessary. It then spawns even more processes (nodes) through launching `rocket_league_train.launch`.

The entire set of environments share the same uuid folder, and each environment creates a plot in the folder using the random port it got as part of the file name. There is a safe race condition for port 11311 (default). This environment will render by default, and is the one you interact with if you run ROS commands in the terminal (without changing `ROS_MASTER_URI`).

Important notes:
- because the main training script is not a ROS node, none of the parameters are accessible. Therefore, there are hardcoded values in the training script for # episodes, # environments, etc
- I moved the save folder to `catkin_ws/data/rocket_league` out of preference. If changed, it has to be in two places unfortunately (yaml file and training script)
- *Adjust the number of environments if you're training on your personal laptop. I suggest a maximum of 4*

## Instructions for use
To train:
- run the docker container as per usual, appending `--gpus all` if desired
- launch the training script via `./rktl_autonomy/scripts/train_rocket_league.py`. **Do not launch it via roslaunch**
- watch it destroy your CPU

To evaluate:
instructions are unchanged. Use the `rocket_league_eval.launch` launch file.

## Testing
The automated tests for the reward function aren't changed. There are no new automated tests for the roslaunch shenanigans, but it has been thoroughly tested manually with up to 64 simultaneous environments.


jrbaxter0:
I need to make sure the tests still work / are testing what we want, and I also need to make the cartpole and snake stuff compatible. I'll have that done by the Wednesday meeting

hmccarty (commented):
This is super cool, everything looks good from a glance. I'll run on my machine tonight.

jrbaxter0:
Assuming the tests don't fail, this should be good to go

halcdev (approved):
All runs on supercomputer properly.

PR #86, opened 2022-02-04
Added an edge detection node to assist with focus so we don't have to keep guessing

ajernandes:


rktl_perception/launch/focus_assist.launch
```diff
@@ -0,0 +1,6 @@
+<launch>
```
dgerblick, line 1: This should be modified to start each camera. That way we don't have to run the full stack just to focus the cameras.

rktl_perception/src/focus_vis_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+    {}
+
+void FocusNode::FocusCallback(const sensor_msgs::ImageConstPtr& msg, const sensor_msgs::CameraInfoConstPtr& info) {
+    try {
+        /* Convert the ROS message into a cv_ptr & dereferencing the pointer to an OpenCV Mat */
+        cv_bridge::CvImagePtr cv_ptr = cv_bridge::toCvCopy(msg, "bgr8");
+        cv::Mat current_frame = cv_ptr->image;
+        cv::Mat frame_gray, frame_blurred, frame_edge;
+        
+        #if false /* shouldn't be necessary on the new computers */
```
dgerblick, line 26: Why is this here? I think we can just remove it

rktl_perception/src/focus_vis.cpp
```diff
@@ -0,0 +1,17 @@
+/* Node to detect the ball's position on the field.
```
dgerblick, line 1: This is false

rktl_perception/src/focus_vis_node.cpp
```diff
@@ -0,0 +1,51 @@
+/* Node to detect the ball's position on the field.
```
dgerblick, line 1: This is also false

rktl_perception/include/rktl_perception/focus_vis_node.h
```diff
@@ -0,0 +1,50 @@
+/* Node to detect the ball's position on the field.
```
dgerblick, line 1: This is very false

rktl_perception/include/rktl_perception/focus_vis_node.h
```diff
@@ -0,0 +1,50 @@
...
+ * License:
+ *   BSD 3-Clause License
+ *   Copyright (c) 2020, Autonomous Robotics Club of Purdue (Purdue ARC)
+ *   All rights reserved.
+ */
+
+#pragma once
```
dgerblick, line 8: This is completely fine and works perfectly but I just don't like it and I'm leaving this comment out of pure spite.

dgerblick (changes requested):
Made a few comments. Most you can ignore, The only one that matters is the launch file one.

rktl_perception/src/focus_vis_node.cpp
ajernandes: I presume you're talking about the resizing, which is gone. if you're talking about the `cv::Mat` stuff, that has to stay.

rktl_perception/src/focus_vis.cpp
ajernandes: mmhmm

rktl_perception/include/rktl_perception/focus_vis_node.h
ajernandes: I'm leaving it for the same reason. idek what it does, James had it in code that I copied and pasted a year ago.

rktl_perception/include/rktl_perception/focus_vis_node.h
ajernandes: mmhmm

rktl_perception/src/focus_vis_node.cpp
ajernandes: k.

rktl_perception/launch/focus_assist.launch
ajernandes: Good point.

rktl_perception/src/focus_vis_node.cpp
dgerblick: I meant the \#if false stuff

ajernandes: gone & pushed

dgerblick (approved):
I rescind all previous comments (except for the pragma once stuff, I'm team include guards until I die)

PR #89, opened 2022-02-05
add launch file tests

jrbaxter0:
This add this feature: http://wiki.ros.org/roslaunch#Catkin

I believe it verifies that the launch files are not broken, and that the `package.xml` lists all required dependencies. Currently, we fail the tests 🥇

Is this something we'd want as part of the test suite? If so, it should be easy to fix the errors.


ajernandes:
well, it didn't get to perception, though `catkin test rktl_perception` did work on my computer.

jrbaxter0:
@ajernandes i think the changes to CMakeLists are not needed. It compiled previously, so it's fine I think. This test just gets upset with package.xml. Perhaps nodelet needs to be an exec depend? Or is it actually used when compiling?

jrbaxter0:
If this runs OK, then I think this is good for merge. Also installs `rosserial-arduino` which is a bug impacting like 3 branches. We had to install it anyways to pass the test. @jaketockerman

ajernandes:
With installing `rosserial-arduino` I was able to get all tests to pass, but I assume that it won't pull the new docker image for tests until it's merged in? Is there a way to get it to run the tests with the new docker image?

jrbaxter0:
It is supposed to use the new docker image. This could indicate a bug in the CI workflow

jrbaxter0:
47d745ac7f023b7ef1ef140d4f9169f607f1db3b should have fixed it.

jrbaxter0:
restarted the tests

jrbaxter0:
This is why it failed:
```
[/home/runner/catkin_ws/src/rocket_league/rktl_autonomy/launch/snake/snake_train.launch]:
	cannot find package [snakesim] for node [snake_node]
```

The CI is fixed wrt pulling using the new Docker container. However, this test will fail if the `arc_tutorial` repo doesn't exist. Do we pull it in the container too, or do we tell it to ignore that file . . .

jrbaxter0:
I opted for the latter. Seemed wrong to have this repo's tests depend on another. We can change that later if someone feels that was the wrong decision

ajernandes (approved):
I'm ok ignoring tests that require a different repo.

PR #90, opened 2022-02-06
arduino-refactor

jaketockerman:
- Steering trim offset for each car
- Throttle rate limit
- Steering flip option
- Enable message


jrbaxter0:
@jaketockerman To get this locally, you can run `./docker-build.sh --build-arg TAG=90` and it should use the container it just built in the cloud. However, that'll take forever to download so you may just want to use `apt` inside your regular container.

docker/Dockerfile
```diff
@@ -47,6 +47,8 @@ RUN pip3 install pybullet stable-baselines3 && \
     apt update && \
     DEBIAN_FRONTEND=noninteractive apt install -y \
     python3-pygame \
+    ros-${ROS_DISTRO}-rosserial-arduino \
+    ros-${ROS_DISTRO}-rosserial \
```
jrbaxter0, line 51: This might be redundant since `rosserial-arduino` should install `rosserial`. Not sure if we really care though, since the command will still work just fine.

jrbaxter0:
@jaketockerman #89 put this in the main Docker repo. You can get it with `./docker-build.sh --pull` once the workflow for 45a52230fb092370d5a3f346bfc2ef686e95a610 finishes.

jrbaxter0:
Worked well, but the parameters didn't do anything. I think I fixed that but need to wait to test before merging

jrbaxter0 (approved):
I'm approving some of my own code ;)

PR #91, opened 2022-02-06
Docker refactor

jrbaxter0:
this might greatly simplify the image. Could also break everything.


jrbaxter0:
image
Certainly seems to reduce the size. I believe this is from not using CUDA as the base? SB3 uses pytorch, which should have CUDA built in, without needing to install separately

jrbaxter0:
image
I think this should work then

jrbaxter0:
Reminder that we need to revert this before merging: d87f732

jrbaxter0:
I changed the branch for the wrong workflow. Should work now

jrbaxter0:
I deleted #91 on the DockerHub, just so it wouldn't get confused if a new image with the same tag only has one platform.

dgerblick (approved):

PR #92, opened 2022-02-07
Control

jrbaxter0:
It's been ages since this code was merged to main. We'll take a look at it in M2, but I wanted to clean it up so there is a starting point.


jrbaxter0:
Actually, we should wait to merge this until there are test cases for the controller. I can do that sometime this week or next.

PR #93, opened 2022-02-08
Sim fixes & test cases

hmccarty:
## Changes

Primary purpose was to refactor test cases. Upon refactoring, motion of car was determined to be inaccurate. Approach moved from controlling car position directly to joint-based velocity control. Appears to be a lot more accurate and match expected curves of 2nd order response.

Also formatted python files using Black and job to check for Black formatting errors.

## Testing

Can run test cases with:
```
catkin test
```

Also able to test manually with keyboard:
```
roslaunch rkt_launch rocket_launch_sim.launch
rosrun rktl_sim keyboard_interface
```


rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -85,8 +101,12 @@ def __init__(self, node_name='gym_interface', eval=False, launch_file=None, laun
...
         # additional set up for training
         if not self.__EVAL_MODE:
-            self.__DELTA_T = rospy.Duration.from_sec(1.0 / rospy.get_param('~rate', 30.0))
-            self.__clock_pub = rospy.Publisher('/clock', Clock, queue_size=1, latch=True)
+            self.__DELTA_T = rospy.Duration.from_sec(
+                1.0 / rospy.get_param("~rate", 30.0)
+            )
+            self.__clock_pub = rospy.Publisher(
+                "/clock", Clock, queue_size=1, latch=True
+            )
```
jrbaxter0, line 109: I'm not sure if I like the auto formatting. Personally, I think this is a step down in this particular instance.

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -139,18 +159,20 @@ def reset(self):
         if self._has_state():
             # generate log
             info = {
-                'episode'    : self.__episode,
-                'net_reward' : self.__net_reward,
-                'duration'   : (rospy.Time.now() - self.__start_time).to_sec()
+                "episode": self.__episode,
+                "net_reward": self.__net_reward,
+                "duration": (rospy.Time.now() - self.__start_time).to_sec(),
             }
```
jrbaxter0, line 165: same here

rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
```diff
@@ -20,38 +20,54 @@
...
+        launch_args=[],
+        run_id=None,
+    ):
+        super().__init__(
+            node_name="rocket_league_agent",
+            eval=eval,
+            launch_file=launch_file,
+            launch_args=launch_args,
+            run_id=run_id,
+        )
```
jrbaxter0, line 40: Here its probably an improvement though

rktl_sim/CMakeLists.txt
```diff
@@ -4,16 +4,25 @@ project(rktl_sim)
...
+  message_generation
 )
 
 # install python module(s)
 catkin_python_setup()
 
+# services
+add_service_files(
+  FILES
+  SetSimEffort.srv
```
jrbaxter0, line 16: What does this do?

rktl_sim/nodes/simulator_node
```diff
@@ -122,21 +124,36 @@ class SimulatorROS(object):
             rospy.signal_shutdown(
                 'no urdf file exists at path {}'.format(param))
 
+    def set_effort(self, effort_req):
```
jrbaxter0, line 127: You are using a service rather than the topic?

jrbaxter0 (commented):
Would like to discuss if we should use the black formatting or not. I think would definitely be nice to standardize everything, and that is probably worth the few cases where it makes things look worse.

I also want to know what exactly that service is doing.

jrbaxter0 (changes requested):
So from our conversation, I am in favor of using `pyflakes` to catch more serious issues rather than using an autoformatter. I also think rather than implementing a service to help with test cases, you should have the test cases interface directly with the sim without ROS.

jrbaxter0 (approved):
If `SPEED_INIT` was removed accidentally, it should be added back in. Otherwise all looks good

PR #97, opened 2022-02-13
hotfixed issue with obs bounds

hmccarty:
Sim was throwing errors about odoms escaping observation bounds. This was caused by a change in the sim, which gave the goal depth.

More work is required to improve parameter layout. Will do in separate PR.

## Testing

Running the autonomy script will no longer produce the out of bounds error.


halcdev (approved):
no more error messages 😃

PR #100, opened 2022-02-28
Field visualizer

jrbaxter0:
Adds a pretty cool visualization system that allows us to project the data from all cameras onto a ground plane. Useful for validating camera calibration and also filtering

# testing
Nothing automated, but it has been heavily tested in practice. To run off a bag file:
```
roscore
rosparam set use_sim_time true
roslaunch rktl_launch field_view.launch
rosbag play <bag> --clock
```


ajernandes (approved):
looks good

PR #101, opened 2022-03-04
Take turns

jrbaxter0:
This mitigates timeout issues caused by launching many environments simultaneously by enforcing them to go one at a time. It allows the supercomputer to run many, many envs at once, with the minor downside of launching taking a bit longer.


jrbaxter0:
Note that this doesn't yet resolve the issues with random processing existing when it closes improperly :(

halcdev (approved):
Works as intended. The more envs added, the slower it is to launch.

PR #102, opened 2022-03-04
Perception readme

ajernandes:
This will need to be updated later, but for now it's good enough


jrbaxter0 (approved):
Glad we got this closer to being up-to-date

PR #103, opened 2022-03-04
Car physics

jrbaxter0:
Changes the physics model to something that should be more accurate. I removed the wall collision logic since it was a hack and we should figure out a different way to do that.

Needs input from @hmccarty on best way to get car's physical values into sim. I assume you don't want a ROS dependency that deep down?


rktl_sim/src/simulator/car.py
```diff
@@ -13,24 +13,20 @@
...
-        self._length = length
-        self._length_r = self._length / 2.
-        self._steering_rate = 2*self._steering_limit / 0.25
+
+        # physical constants
+        self._LENGTH = length
+        self._MAX_SPEED = 1.0
+        self._THROTTLE_TAU = 0.25
+        self._STEERING_THROW = math.pi / 6.
+        self._STEERING_RATE = 2*self._STEERING_THROW / 0.25
```
jrbaxter0, line 22: These 5 lines are what I mean. They should all be synced to single rosparam file somewhere that the filters, controllers, and sim will all use.

hmccarty: That will done in root launch config and passed through node

hmccarty (changes requested):
You should modify test cases to fit new model

rktl_sim/src/simulator/sim.py
```diff
@@ -181,25 +181,19 @@ def step(self, throttle_cmd, steering_cmd, dt):
...
-                if contact[2] in self._walls:
-                    wallContact = True
-                    break
-            car.step((throttle_cmd, steering_cmd), wallContact, dt)
-    
         # PyBullet steps at 240hz
-        for _ in range(int(dt * 240.)):
+        p_dt = 1.0 / 240.0
+        for _ in range(round(dt / p_dt)):
+            # Step kinematic objects independently, at max possible rate
```
jrbaxter0, line 187: Everything is ready to be merged. Only possible hiccup is this. The bicycle model stuff does greatly benefit from being run at a rate higher than 10 Hz, but 240 might be excessive.

hmccarty: See `setTimeStep` in the [quickstart guide](https://usermanual.wiki/Document/PyBullet20Quickstart20Guide.543993445/html)

rktl_sim/nodes/simulator_node
```diff
@@ -83,9 +83,16 @@ class SimulatorROS(object):
...
+        car_properties['length'] = rospy.get_param('/cars/length')
+        car_properties['max_speed'] = rospy.get_param('/cars/throttle/max_speed')
+        car_properties['throttle_tau'] = rospy.get_param('/cars/throttle/tau')
+        car_properties['steering_throw'] = rospy.get_param('/cars/steering/max_throw')
+        car_properties['steering_rate'] = rospy.get_param('/cars/steering/rate')
+
         # Creating physics simulator
         self.sim = simulator.Sim(
-            urdf_paths, field_setup, spawn_bounds, speed_init, render_enabled)
+            urdf_paths, field_setup, spawn_bounds, speed_init, render_enabled, car_properties)
```
hmccarty, line 95: Params for sim are starting to get a little crazy. In the future, maybe we considering adding a factory class.

hmccarty (approved):
Everything looks good. Comments above were more for discussion.

rktl_sim/nodes/simulator_node
jrbaxter0: I'll leave that to you

rktl_sim/src/simulator/sim.py
jrbaxter0: If PyBullet is best tuned for 240Hz, I probably don't want to change that. Instead we could look at instead maybe stepping these at 30 or 60 Hz inside that loop. But I figure we only do that if we realize it is slowing stuff down.

PR #104, opened 2022-03-05
Network refactor

jrbaxter0:


rktl_control/nodes/velocity_controller
```diff
@@ -61,7 +61,7 @@ class VelocityControllerROS(object):
         self.steering_pub = rospy.Publisher('effort/steering', Float32, queue_size=1)
 
         # Subscribers
-        rospy.Subscriber('cmd_vel', Twist, self.command_cb)
+        rospy.Subscriber('cmd', Twist, self.command_cb)
```
jrbaxter0, line 64: can we use `control` instead? This also shouldn't be a twist. We need to make a custom message for it.

rktl_control/launch/mean_odom_filter.launch
```diff
@@ -0,0 +1,5 @@
+<launch>
+  <node pkg="rktl_control" type="mean_odom_filter" name="filter" output="screen">
+    <rosparam command="load" file="$(find rktl_control)/config/mean_odom_filter.yaml" />
+  </node>
+</launch>
```
jrbaxter0, line 5: Rather than having many nearly empty launch files, I think we should do something like what perception did and have a launch file that runs the whole stack:

- particle or odom filter
- velocity controller

I guess the ball uses just a filter. But rather than `mean_odom_filter.yaml`, we should have a ball and a car one.

rktl_launch/launch/car.launch
```diff
@@ -0,0 +1,17 @@
...
+
+        <group if="$(arg real)">
+            <!-- Control -->
+            <include file="$(find rktl_control)/launch/velocity_controller.launch" />
+
+            <!-- Hardware -->
+            <include file="$(find rktl_control)/launch/hardware_interface.launch" />
+        </group>
+    </group>
+</launch>
```
jrbaxter0, line 17: Like this. Given that all these files are in `rktl_control`, I think this launch file should exist there. That way control can be modified independently.

jrbaxter0 (commented):
Because I'm the author, it won't let me request changes

ajernandes (commented):
I am not sure if I did the calibrator node correct. It may be something that we only know when we actually test it.

Other than that, it does try to launch the correct camera, so I assume it works

ajernandes (approved):
Everything in perception looks good to me!
Everything else looks good too, but I'm less qualified to speak on that...

jrbaxter0 (commented):
Everything looks good to me, but I also authored a lot of these changes. I'd like to merge this in tonight since we have several other branches that depend on these changes.

PR #105, opened 2022-03-05
Bundles

ajernandes:
this is ready to be merged


rktl_perception/launch/camera_boilerplate.launch
```diff
@@ -23,25 +23,25 @@
...
-        <param name="origin_id" type="str" value="0"/>
+        <param name="origin_id" type="str" value="0,1,2,3,4,5,6,7"/>
         <param name="buffer_size" type="int" value="300"/>
         <param name="ball_sub_topic" type="str" value="ball_vec"/>
         <param name="ball_pub_topic" type="str" value="/ball/pose"/>
         <param name="ball_radius" type="double" value="0.05"/>
         <rosparam param="pub_topics">
-      '0': /origin/pose
-      '2': /car0/pose
+            '0,1,2,3,4,5,6,7': /origin/pose
```
jrbaxter0, line 32: This is publishing many different tags on the same topic? Seems like we'd need to publish them separately so the localizer can use them.

jrbaxter0 (approved):

PR #108, opened 2022-03-08
Control

jrbaxter0:
This completely redoes the controller node, and adds a new message `ControlCommand`. The idea is that the bezier curve stuff spits out an ideal velocity and curvature, and the controller handles converting it to steering and throttle. Previously, you gave an ideal throttle and steering directly, which is asking producers to know more about the car than they probably do.

There are no test cases currently, but they will be added when the controller tuning takes place


hmccarty (approved):
Not the biggest fan of the name `ControlCommand`, but everything seems right. Will test correctness on my branch.

PR #109, opened 2022-03-08
Filtering

jrbaxter0:
This adds a super cool particle filter for the car. There is a test case to ensure the bicycle model is properly implemented


ajernandes (changes requested):
Can you add a readme & more comments explaining what this does & how it works so we can maintain it after you leave

ajernandes (approved):
Thank you for adding documentation

PR #111, opened 2022-03-11
Path planning / Additional sim features

hmccarty:
# Changes

- Integrated prior path following code to work with Dan's refactored path planning.
- Added Gaussian sensor noise
- Added option to perfectly follow velocity and curvature commands

# Testing

No official test cases yet (see future section).

Any manual testing can be done via the main sim launch file:
```
roslaunch rktl_launch rocket_league_sim.launch
```
For path planning, use rqt or command line to call the `reset_planner` service.

Testing the sim features can be done via roslaunch arguments.

`simulate_car_effort` determines whether the car perfectly follows commands (true for realistic, false for perfect).

`simulate_filter` determines whether noisy sensor data is published. (true for publishing noisy data to `sync_pose`, false for publishing perfect data to `odom`)

# Future

Few more things I would like to do either in this PR or future PR:
- Added a few baseline test cases for path planning / following
- Ensure sim and planner packages conform to a standard style (maybe create a rktl style guide?)
- Fix a few bugs with path planning / following
- Continue tuning path planning / following


rktl_planner/launch/simple_agent.launch
```diff
@@ -0,0 +1,19 @@
+<launch>
+    <arg name="agent_name" default="agent0"/>
+    <arg name="car_name" default="car0"/>
+
+    <group ns="agents/$(arg agent_name)">
```
jrbaxter0, line 5: I think we should just put the autonomy stuff in the same namespace as the cars. Would make things much simpler for topics

rktl_planner/launch/simple_agent.launch
```diff
@@ -0,0 +1,19 @@
...
+        <!-- Path Follower -->
+        <node pkg="rktl_planner" type="path_follower" name="path_" output="screen">
+            <rosparam command="load" file="$(find rktl_planner)/config/path_follower.yaml" />
+            <param name="car_name" value="$(arg car_name)" />
```
jrbaxter0, line 9: I dislike putting a space at the end here. Whichever way we go, I think it would be worth standardizing.

jrbaxter0 (changes requested):
The car "vibrates" in the sim. I assume this is from you publishing Odometry with artificial noise? I think that should only happen if we are looking to test the filter. If we want a simplified environment for sim testing, I think we should publish true values. I especially think the viz should be rendering the true value.

rktl_planner/launch/simple_agent.launch
hmccarty: The plan was to separate `autonomy` to avoid ruling out a multi-car agent. It's also nice to separate, because all `car` topics and nodes can be placed in a single launch file under `rktl_control` while `autonomy` or `agent` topics are all bundled in their own launch file.

jrbaxter0: I think it is nice because rather than doing:
```
        self.bot_velocity_cmd = rospy.Publisher(f'/cars/{car_name}/command',
            ControlCommand, queue_size=1)

        # Subscribers
        rospy.Subscriber(f'/cars/{car_name}/odom', Odometry, self.odom_cb)
```

We can just directly subscribe to `command` and `odom` in the current namespace. I'm a big fan of not hardcoding anything in the Python source, and using remaps / namespacing in the launch files.

However, if you feel strongly about this, I don't think the existing system is wrong. I just think the other way would be cleaner.

jrbaxter0: At the same time, there are so many nodes and things going on in there, maybe we just leave it as is since it nicely spaces things. I didn't realize it was more than 2 nodes and topics. For the neural network one, I'd like to put it in car since it is just one node.

jrbaxter0:
@hmccarty Can you add a README for `rktl_planner`? It should at the very least have a basic description of the nodes you've made and what they do, and commands on how to tell the agent to score a goal after running `roslaunch rktl_launch rocket_league_sim.launch sim_mode:=ideal`. For this PR, I do not want to worry about issues with the controller or bugs that happen when realistic. We can fix those in the upcoming controls PR. We can also make "ghost mode" ground truths in a new PR as well.

rktl_planner/CMakeLists.txt
```diff
@@ -3,13 +3,32 @@ project(rktl_planner)
...
 ## Uncomment this if the package has a setup.py. This macro ensures
 ## modules and global scripts declared therein get installed
 ## See http://ros.org/doc/api/catkin/html/user_guide/setup_dot_py.html
 catkin_python_setup()
 
+# msg and srv
+add_service_files(
+  FILES
+  CreateBezierPath.srv
+)
```
jrbaxter0, line 23: We have the `rktl_msgs` package for a reason. I think it makes sense to stick this in there.

rktl_planner/src/rktl_planner/pure_pursuit.py
```diff
@@ -86,6 +86,11 @@ def calculate_angle(intersect_pos, bot_pos, bot_orient, lookahead_dist, bkw):
     dist = np.linalg.norm(tang_line)
     angle = (2 * math.asin(round((dist / 2) / lookahead_dist, 6)))
     sign = np.sign(np.cross(intersect_pos - bot_line, bot_line))[2]
+
+    # if angle < 0:
+    #     angle += math.pi
+    # else:
+    #     angle -= math.pi
```
jrbaxter0, line 93: Is this needed? If not, better to delete it to not confuse.

rktl_sim/config/simulation.yaml
```diff
@@ -1,4 +1,18 @@
...
+  #   pos: [0.5, 0.0, 0.06]
+  sensor_noise:
+    pos: [0.01, 0.01, 0.00]
+    orient: [0.0, 0.0, 0.01]
+
+ball:
+  init_speed: 0.0
+  # init_pose:
+  #   pos: [0.0, 0.5, 0.06]
+  #   orient: [0.0, 0.0, 0.0]
```
jrbaxter0, line 15: are these parameters valid / in use? If they are commented for a reason, pls explain why in the file.

jrbaxter0 (changes requested):
When doing tests, the agent seemed pretty fragile. I'd really like to make it not spit out errors at 10Hz when it fails. Instead, maybe it can error once and give up.

I also pointed out a few code style things.

jrbaxter0:
Also, the tests definitely need to be fixed before merging.

rktl_planner/CMakeLists.txt
hmccarty: That is a private service only being used by nodes in the same package. Which is why we didn't add it to the global msgs package.

rktl_sim/config/simulation.yaml
hmccarty: Valid parameters, I left it there as an example

jrbaxter0 (approved):

jrbaxter0:
leaving the branch b/c I assume you're using it to improve the path planning algos

PR #113, opened 2022-03-27
Remote x11 forwarding

jaketockerman:
use --net=host on server and ssh -Y to pass x11 forwarding over ssh

### DIFF
--- a/docker/docker-run.sh
+++ b/docker/docker-run.sh
@@ -24,5 +24,5 @@ docker run --rm \
     -v ~/.ssh:/home/$USER/.ssh:ro \
     --name $USER-$CONTAINER_NAME \
     $@ \
-    $REPO_NAME:local \
+    $REPO_NAME:local-$USER \
     ${DOCKER_CMD:+/bin/zsh -c "$DOCKER_CMD"}
--- a/docker/docker-run.sh
+++ b/docker/docker-run.sh
@@ -20,7 +20,6 @@ docker run --rm \
     -v $WS_DIR:/home/$USER/catkin_ws \
     -v /tmp/.X11-unix:/tmp/.X11-unix \
     `[ -f ~/.gitconfig ] && echo "-v $HOME/.gitconfig:/home/$USER/.gitconfig:ro"` \
-    -v $HOME/.Xauthority:/home/$USER/.Xauthority \
     -v ~/.ssh:/home/$USER/.ssh:ro \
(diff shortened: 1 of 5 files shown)


jrbaxter0 (approved):
As long as you provided **detailed** instructions to members over Discord, this is fine with me. I have not reviewed the server functionality at all since that is not my domain. When this is merged, I expect there to be an @rocketleague announcement on #rl-general with the exact copy-paste command to rename the image.

jrbaxter0:
But as a comment, if you use `--net host` when running on the server, only one person can be running ROS stuff at a time by default. It will use port 11311 and get upset that there is already something running. You can change the default port, which is something I do for training hundreds of simultaneous environments on the super computer. But that code is somewhat flakey.

Probably you want to bind only the port actually needed for display.

jrbaxter0:
It also might be better to use this solution so that each user's docker images / containers are specific to themselves. Then you don't need to worry about people killing other's stuff, and doing this tagging stuff: https://forums.docker.com/t/share-docker-between-users/5728

jrbaxter0:
But again, whatever solution you decide on for the server is fine as long as it doesn't break the repo.

docker/docker-run.sh
```diff
@@ -20,8 +20,9 @@ docker run --rm \
     -v $WS_DIR:/home/$USER/catkin_ws \
     -v /tmp/.X11-unix:/tmp/.X11-unix \
     `[ -f ~/.gitconfig ] && echo "-v $HOME/.gitconfig:/home/$USER/.gitconfig:ro"` \
+    -v $HOME/.Xauthority:/home/$USER/.Xauthority \
```
jrbaxter0, line 23: This is not portable to Mac. We already do this exact thing on line 19, but do it as read only, which I'd prefer unless if there is a need to be able to change it.

jrbaxter0 (changes requested):
My review is primarily concerned with "will this break the project?" as opposed to "will this work on the server?"

That being said, my two major issues are with the comments in made in `docker-run.sh`. All the other comments are just suggestions.

I also think we should update the Docker README and link to it in the main README like we do other sections.

hmccarty:
Works for me

jrbaxter0 (approved):
Changes to `-build`, `-join`, `-run` look good to me. Will not mess up any existing workflows if containers are rebuilt when merged.

PR #115, opened 2022-03-28
cleaned up changes that I hacked in yesterday

ajernandes:


rktl_perception/launch/camera.launch
```diff
@@ -50,12 +50,13 @@
...
-            <param name="min_vib" value="30" />
+            <param name="min_hue" value="50" />
             <param name="max_hue" value="100" />
-            <param name="max_sat" value="200" />
-            <param name="max_vib" value="80" />
+            <param name="min_sat" value="70" />
+            <param name="max_sat" value="180" />
+            <param name="min_vib" value="40" />
+            <param name="max_vib" value="100" />
+            <param name="min_size" value="50" />
```
jrbaxter0, line 59: I think it would be worth putting these in a yaml file

rktl_perception/src/ball_detection_node.cpp
```diff
@@ -39,11 +40,6 @@ void BallDetection::BallCallback(const sensor_msgs::ImageConstPtr& msg, const se
         cv_bridge::CvImagePtr cv_ptr = cv_bridge::toCvCopy(msg, "bgr8");
         cv::Mat current_frame = cv_ptr->image;
         cv::Mat frame_HSV, frame_threshold;
-        
-        #if false /* shouldn't be necessary on the new computers */
-        /* resize image for performance */
-        cv::resize(current_frame, current_frame, cv::Size(), 0.5, 0.5);
-        #endif
```
jrbaxter0, line 46: lmao. This is like commented out code on steroids.

ajernandes:
> From my perspective, it seemed like the ball was still being located in a spot where it wasn't actually. ... Do we know where the issues is?

I don't have a solid answer, but I have a theory. My theory is that my threshold for the size was too low which caused other cameras to see noise and think it's the ball, so all of those positions were getting averaged together. My hope is tuning the size threshold will fix this.

jrbaxter0:
> > From my perspective, it seemed like the ball was still being located in a spot where it wasn't actually. ... Do we know where the issues is?
> 
> I don't have a solid answer, but I have a theory. My theory is that my threshold for the size was too low which caused other cameras to see noise and think it's the ball, so all of those positions were getting averaged together. My hope is tuning the size threshold will fix this.

Ok, seems like a reasonable hypothesis. If that doesn't work, there is a hack that might. You publish each pixel individually, and since the pose synchronizer takes the average reading within a window, it'll put the mean pretty close to the actual ball position.

Or, we somehow assign weights (based off of how big the blob is). The only way to use that weight might be to publish it multiple times to use the averaging code. That or we use covariance like we actually should, but that will require a lot of modifications in a small amount of time.

I guess both of those are useful if there isn't a clear cutoff between blob size for noise and real measurements.

ajernandes:
> > > From my perspective, it seemed like the ball was still being located in a spot where it wasn't actually. ... Do we know where the issues is?
> > 
> > 
> > I don't have a solid answer, but I have a theory. My theory is that my threshold for the size was too low which caused other cameras to see noise and think it's the ball, so all of those positions were getting averaged together. My hope is tuning the size threshold will fix this.
> 
> Ok, seems like a reasonable hypothesis. If that doesn't work, there is a hack that might. You publish each pixel individually, and since the pose synchronizer takes the average reading within a window, it'll put the mean pretty close to the actual ball position.
> 
> Or, we somehow assign weights (based off of how big the blob is). The only way to use that weight might be to publish it multiple times to use the averaging code. That or we use covariance like we actually should, but that will require a lot of modifications in a small amount of time.
> 
> I guess both of those are useful if there isn't a clear cutoff between blob size for noise and real measurements.

There is a major difference, the noise is usually just a blob of a few pixels, while the ball is significantly larger. I’m confident that I can have that sorted for the next test. I also tried the code on the bag from last Wednesday and when it wasn’t putting the ball on the tape, it was spot on.

rktl_perception/src/ball_detection_node.cpp
ajernandes: there is a reason it's gone

rktl_perception/launch/camera.launch
ajernandes: Made this change

ajernandes:
Are there any other changes you wanted to make? The value for `min_size` will have to be tuned, but I can't do that until next meeting.

jrbaxter0:
> Are there any other changes you wanted to make? The value for `min_size` will have to be tuned, but I can't do that until next meeting.

As far as code quality goes, I think this all looks good. If you want to add a test cases, I always think that is good to do. But otherwise, I think it is good to be merged in.

ajernandes:
> If you want to add a test cases, I always think that is good to do.

One day I will, that day is not today.

PR #116, opened 2022-03-28
Training updates

jrbaxter0:
This does the following:
- changes the action space from box to discrete. This is backed up with evidence that it made it much better at hitting the ball
- adds a script to allow running several training jobs at once. The idea is that you can run experiments now with defined variables easily in order to narrow down on an agent that will do the trick.
- adds a comprehensive readme to autonomy. This should thoroughly explain all the new features


halcdev (approved):
We tested it before so 👍

PR #117, opened 2022-03-29
Control

jrbaxter0:
This adds a functional controller to the realistic simulation environment. Try it out by launching the simulator in realistic mode, then manually publishing ControlCommand messages.

There is likely more that can be done to make the controller as good as possible. I'll likely experiment with root locus methods if this isn't satisfactory when running on the real hardware.


rktl_control/nodes/controller
```diff
@@ -136,6 +136,9 @@ class Controller(object):
 
     def odom_cb(self, odom_msg):
         """Callback for odom messages from car."""
+        if self.vr_ref is None or self.psi_ref is None:
```
hmccarty, line 139: Is this common? Do we want to throw a warning, just so that we aren't oblivious to a bunch of messages being dropped.

hmccarty (commented):
Overall looks good.

rktl_control/nodes/controller
jrbaxter0: This is just so that the controller is disabled until we give it an input. Otherwise it'll bounce around holding zero velocity when we first turn it on and the filters are still initializing.

hmccarty: Oh I see. When I first skimmed through I thought you were referencing variables in the msg, but this makes more sense.

hmccarty (approved):

PR #119, opened 2022-04-02
Added option to direct ball at object

hmccarty:
## Changes

Parameterized ball initial speed direction.

## Testing

Uncomment `direct_at_object` in `simulation.yaml`, then launch sim as per usual.

For `direct_at_object`, you can select any URDF defined in `field_setup`. Can document this more later. Touching this code made me realize that some overall refactoring / documentation is required, but I will introduce this in a future PR to avoid breaking changes.

### DIFF
--- a/rktl_sim/config/simulation.yaml
+++ b/rktl_sim/config/simulation.yaml
@@ -10,6 +10,7 @@ car:
 
 ball:
   init_speed: 0.0
+  # direct_at_object: 'goalB' # (goalA, goalB, etc.)
   # init_pose:
   #   pos: [0.0, 0.5, 0.06]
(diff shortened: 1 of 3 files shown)


hmccarty:
One second, a planner commit got mixed in. I need to take that out.

jrbaxter0:
Is it possible to have it not always shoot for dead center goal? I think the car will just learn to park itself right in front of the goal. I think it would be better if it had to sometimes do center, sometimes do sides, sometimes it would bounce off the wall and go in. I'm just not sure how much it would learn.

That being said, I'll spin up an environment on this branch and see what it learns to do. But I don't think we should merge it into main yet since as you said it needs to be refactored already and this is just adding to a fire.

PR #120, opened 2022-04-03
Hardware interface

jrbaxter0:
This refactors the Arduino code and also adds in a safety kill switch. Additionally, it will reload parameters without needing to kill and restart the entire MCU. It also lets it run the first go, eliminating the previous bug with relaunching.


jrbaxter0:
I'm going to merge this as an admin because it works.

PR #128, opened 2022-04-12
Eval script

jrbaxter0:
This is the script I used to make those Excel plots. All detail to run is in README.


halcdev (approved):
Logging to CSV file works, it also creates the new folder directory as intended (right location, right name). README was also very helpful. Maybe implement an auto plotting feature next so that we don't have to do it in Matlab? 😃

PR #129, opened 2022-04-13
Bundle calibration

jrbaxter0:
This adds a matlab script (taken from here: https://github.com/AprilRobotics/apriltag_ros) that can calibrate the field bundle. It has been modified since the original script requires a library that only runs on linux and hasn't been maintained in over a decade. This modifies it to use `rostopic` to convert the bag file to a CSV, then it parses that. Now it can run on any MATLAB instance.


ajernandes:
Do we want to add the new tag YAML in with this PR?

I would also love to fix pyflakes :)

ajernandes (approved):

PR #131, opened 2022-04-14
Merge game manager into main

ajernandes:
This includes code that tracks scoring, manages a clock, provides a web UI, and publishes the "gameStatus" message


rktl_game/CMakeLists.txt
```diff
@@ -0,0 +1,206 @@
...
+## Find catkin macros and libraries
+## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
+## is used, also find other catkin packages
+find_package(catkin REQUIRED COMPONENTS
+  roscpp
+  rospy
+  std_msgs
+)
+
+## System dependencies are found with CMake's conventions
```
jrbaxter0, line 16: I prefer deleting all the stuff that we aren't going to use to make this more readable.

rktl_game/index.html
```diff
@@ -0,0 +1,102 @@
+<!DOCTYPE html>
```
jrbaxter0, line 1: Can we stick all the web stuff in a folder for better organization? Maybe just a folder called `web`?

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,104 @@
...
+  All rights reserved.
+"""
+
+
+
+# ROS
+import rospy
+from std_msgs.msg import Bool, UInt16, Int16, Int8, Header
+from std_srvs.srv import Empty, EmptyResponse, Trigger, TriggerResponse
+from nav_msgs.msg import Odometry
```
jrbaxter0, line 14: obv, fix all the pyflake stuff

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
+"""Use ball location to determine when a goal is scored.
+License:
+  BSD 3-Clause License
+  Copyright (c) 2020, Autonomous Robotics Club of Purdue (Purdue ARC)
```
jrbaxter0, line 5: wrong year

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,104 @@
...
+        rospy.Subscriber('cars/enable', Bool, self.enable)
+
+        # Services
+        rospy.Service('reset_game', Empty, self.reset)
+        rospy.Service('unpause_game', Empty, self.unpause)
+        rospy.Service('pause_game', Empty, self.pause)
+
+        # Counts loops and decrements game clock every second
+        self.timer_ctr = 0
+        # self.unpause(self)
```
jrbaxter0, line 45: I dislike commented out code

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,104 @@
...
+        # main loop
+        rate = rospy.Rate(10) #10hz
```
jrbaxter0, line 47: even though 10Hz is pretty set in stone, it would be nice for this to be in a config file and be grabbed from param server

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,104 @@
...
+        self.match_status = 0
+        self.active_pub.publish(False)
+        return EmptyResponse()
+
+    def reset(self, _):
+        self.orange_score = 0
+        self.blue_score = 0
+        self.enabled = False
+        self.match_status = 0
+        self.game_clock = rospy.get_param('/game_length', 90)
```
jrbaxter0, line 77: Since only your node consumes this info, I think it can be a private param

rktl_game/package.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>rktl_game</name>
+  <version>0.0.0</version>
+  <description>Perception for ARC Rocket League project</description>
```
jrbaxter0, line 5: Description is wrong?

rktl_msgs/msg/MatchStatus.msg
```diff
@@ -1,4 +1,4 @@
-Header header
+# Header header
```
jrbaxter0, line 1: Something might use this, so let's not remove it.

jrbaxter0 (changes requested):
A lot of my stuff is more organizational. If you say it works, I trust you, and we can test it today.

rktl_game/nodes/game_manager
ajernandes: copy paste

rktl_game/CMakeLists.txt
ajernandes: done

rktl_game/index.html
ajernandes: done

rktl_game/nodes/game_manager
ajernandes: done

rktl_game/nodes/game_manager
ajernandes: done

rktl_game/package.xml
ajernandes: done

rktl_msgs/msg/Score.msg
```diff
@@ -0,0 +1,2 @@
+int32 blue
+int32 orange
```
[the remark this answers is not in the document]
ajernandes: probably, but I like the modularity

rktl_game/nodes/game_manager
ajernandes: The webpage uses it

rktl_game/nodes/game_manager
ajernandes: the parameter exists now but I think that can a "summer" project

rktl_msgs/msg/MatchStatus.msg
ajernandes: OK

ajernandes:
Before this can be merged, we will have to ensure that we set the correct IP/URL for the websocket server. I assume it will be something like `r2d2.local` but not sure atm

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+"""Use ball location to determine when a goal is scored.
+License:
+  BSD 3-Clause License
+  Copyright (c) 2010, Autonomous Robotics Club of Purdue (Purdue ARC)
```
jrbaxter0, line 5: Lol

rktl_game/web/index.js
```diff
@@ -0,0 +1,79 @@
+// Connecting to ROS
+// -----------------
+
+var ros = new ROSLIB.Ros({
+    url: 'ws://r2d2.local:9090'
```
jrbaxter0, line 5: Is there a way to make this run on any computer? Like `localhost:9090`?

rktl_game/web/index.js
```diff
@@ -0,0 +1,79 @@
...
+    ros: ros,
+    name: 'reset_game',
+    serviceType: 'std_srvs/Empty'
+});
+
+// Game clock parameters
+
+var game_time_param = new ROSLIB.Param({
+    ros: ros,
+    name: 'game_length'
```
jrbaxter0, line 53: This is the shared parameter? Should it be set in a config file? I don't see a default value nor a value set otherwise.

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self):
+        rospy.init_node('score_keeper')
+
+        self.orange_score = 0
+        self.blue_score = 0
+        self.orange_goal = (rospy.get_param('/field/length', 1) + 0.1) / 2 
+        self.blue_goal = (rospy.get_param('/field/length', 1)  + 0.1) / -2
+        self.enabled = False
+        self.match_status = 0
+        self.game_clock = rospy.get_param('game_length', 90)
```
jrbaxter0, line 28: I think we should talk in person about what is going on with the parameters here. They should probably be private (prefixed by `~`), or absolute (prefixed by `/`). Unless if you are namespacing this?

jrbaxter0 (changes requested):
I'm still confused about the parameter stuff. I also wonder if it is possible to not hardcode it for r2d2 to run.

rktl_game/web/index.js
ajernandes: localhost did not work with `--net host` or `-p 9090`

ajernandes: and if we want it to be viewed on a different device (like a phone) `localhost` wouldn't work

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,108 @@
...
+from rktl_msgs.msg import Score, MatchStatus
+from std_srvs.srv import Empty, EmptyResponse
+
+class ScoreKeeper():
+    def __init__(self):
+        rospy.init_node('score_keeper')
+
+        self.orange_score = 0
+        self.blue_score = 0
+        self.orange_goal = (rospy.get_param('/field/length', 1) - 0.15) / 2
```
jrbaxter0, line 24: What is going on here? Also, it is preferred to not pass in a default for our global params so that it throws an error if it isn't found. This is mainly because the launch file that launches the global params is pretty far from the one that launches the node, so it is easy to accidentally forget to launch them

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self):
+        rospy.init_node('score_keeper')
+
+        self.orange_score = 0
+        self.blue_score = 0
+        self.orange_goal = (rospy.get_param('/field/length', 1) - 0.15) / 2 
+        self.blue_goal = (rospy.get_param('/field/length', 1)  - 0.15) / -2
+        self.enabled = False
+        self.match_status = 0
+        self.game_clock = rospy.get_param('/game_length', 90)
```
jrbaxter0, line 28: I don't think this needs to be global

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,108 @@
...
+        self.manager_rate = rospy.get_param('manager_rate', 10)
```
jrbaxter0, line 29: if this is only ever referenced here, you can just make it a local rather than a member. It doesn't hurt anything to keep it member though. That is just my preferred style.

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,108 @@
...
+                if self.timer_ctr == self.manager_rate:
+                    self.timer_ctr = 0
+                    self.game_clock -= 1
+            self.update_status()
+            rate.sleep()
+
+    def update_status(self):
+        if (self.match_status == 1 and self.game_clock == 0):
+            if (self.orange_score > self.blue_score):
+                self.match_status = 3
```
jrbaxter0, line 60: Hard coded constants are the devil's work. You can reference them using the enum
`self.match_status = MathStatus.VICTORY_TEAM_A`

These hardcoded constants appear in many locations, so make sure you get them all

rktl_game/nodes/game_manager
```diff
@@ -0,0 +1,108 @@
...
+        self.match_status = 0
+        self.active_pub.publish(False)
+        return EmptyResponse()
+
+    def reset(self, _):
+        self.orange_score = 0
+        self.blue_score = 0
+        self.enabled = False
+        self.match_status = 0
+        self.game_clock = rospy.get_param('/game_length', 90)
```
jrbaxter0, line 79: Rather than fetching this param twice, and having duplicated code and default values, I'd save it to a member variable like `self.GAME_LENGTH`. Or maybe fetch it without the default value in both locations to enforce that it needs to be in a config file.

jrbaxter0 (changes requested):
My major concerns are the hardcoded constants and not following our style for using the parameter server. I did not check functionality or any of the stuff in `web`.

rktl_game/nodes/game_manager
ajernandes: I removed the defaults, idk what you didn't know what was going on but if it's why i'm subtracting 15 centimeters it's because the field is never put exactly in the right location so we need the ball to be outside the field by a few inches to keep it from calling a bounce off the wall a goal

jrbaxter0: I'd make the 0.15 a parameter rather than hardcoding. Something private. But yes, that is what I was asking.

rktl_game/nodes/game_manager
ajernandes: It's fetched when you reset in case it as been modified by the webUI. It's not the prettiest solution, but it works for now

rktl_game/nodes/game_manager
jrbaxter0: I meant global as in parameters. Like we don't need `/game_length`, `~game_length` would be fine. No other nodes should need it. If you change it, do it in both places since the code is duplicated

rktl_game/nodes/game_manager
jrbaxter0: I was referring to `self.manager_rate`

rktl_game/web/index.js
```diff
@@ -0,0 +1,79 @@
...
+    ros: ros,
+    name: 'reset_game',
+    serviceType: 'std_srvs/Empty'
+});
+
+// Game clock parameters
+
+var game_time_param = new ROSLIB.Param({
+    ros: ros,
+    name: '/game_length'
```
jrbaxter0, line 53: I'm upset that this has to be global, but I'll deal with it.

jrbaxter0 (approved):

PR #134, opened 2022-04-16
Added support for multiple cars in sim

hmccarty:
## Changes

- Modified sim and viz to allow for any number of cars
- Condensed ground truth and realistic viz into a single window
- Refactored launch files to allow for easy creation of new cars
- Changed variable and function names to conform to rktl standard style

## Testing

You can launch the official launch file `rocket_league.launch`.

Then add the car in sim (kind of like setting a new car on the field):
```
rosservice /sim_create_car "name: 'car1'
init_pos:
  x: 0.0
  y: 0.0
  z: 0.0
init_yaw: 0.0
reset_to_random_pose: true"
```

Then add the car to the network:
```
roslaunch rktl_launch car.launch car_id:=1
```

To see the new car in the visualizer:
```
rosservice /visualizer_refresh "{}"
```

## Future work for sim & viz + misc. updates

As of now, cars are launched with an agent type of either `none`, `planner` or `autonomy`. But I was thinking we could change that to `control_type`, then easily launch xbox and keyboard interfaces in the same launch file. With the options being: `none`, `bezier_agent`, `autonomy_agent`, `xbox`, or `keyboard`.

I started exploring a few different domain adaptation and randomization methods. Which could help detect issues in planners early and improve the usefulness of autonomy agents. So hopefully I can propagate some of those changes into the sim soon.

Going to push another PR today or tomorrow with additional bezier planner changes.


dgerblick:
Made redundant by #146

PR #137, opened 2022-04-19
Patrol planner

jrbaxter0:
A simple, configurable planner that has a chance at working. Tuning will be required on the hardware, but it works well enough in sim

### DIFF
--- a/rktl_planner/nodes/patrol_planner
+++ b/rktl_planner/nodes/patrol_planner
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""Contains the Pendulum node, a very simple planner.
+"""Contains the PatrolPlanner node, a very simple planner.
 License:
   BSD 3-Clause License
   Copyright (c) 2022, Autonomous Robotics Club of Purdue (Purdue ARC)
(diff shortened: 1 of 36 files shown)


rktl_planner/nodes/patrol_planner
```diff
@@ -0,0 +1,266 @@
...
+        NORTH_HEADING_ERROR = sad(yaw, 0.0)
+        SOUTH_HEADING_ERROR = sad(yaw, math.pi)
+        EAST_HEADING_ERROR = sad(yaw, -math.pi/2.0)
+        WEST_HEADING_ERROR = sad(yaw, +math.pi/2.0)
+
+        # Prep command message for publish
+        cmd_msg = ControlCommand()
+
+        # check to see if an existing attempt has timed out
+        if self.attempt_start_time is not None and (rospy.Time.now() - self.attempt_start_time).to_sec() > self.ATTEMPT_TIMEOUT:
```
hmccarty, line 104: Picky, but maybe break into 2 lines?

rktl_planner/nodes/patrol_planner
```diff
@@ -0,0 +1,266 @@
...
+            return
+
+        # extract ball position and velocity, adding look-ahead time
+        bx, by = self.ball_position
+        vbx, vby = self.ball_velocity
+        bx += vbx * self.LOOKAHEAD_TIME
+        by += vby * self.LOOKAHEAD_TIME
+
+        # check if we should be offensive or defensive
+        if bx > 0.0:    # offensive
```
hmccarty, line 123: Maybe we look into adding a slight offset. Like we may not want to be offensive until the ball is in the last 1/4th of the field.

hmccarty: At least might be something to parametrize

hmccarty (changes requested):
Most of the comments are style-focused since we've already seen it work fairly well irl.

rktl_planner/nodes/patrol_planner
jrbaxter0: done

hmccarty (approved):

PR #138, opened 2022-04-19
Physical properties

jrbaxter0:
This updates physical properties based off of what @Dhirodaatto and I measured today

### DIFF
--- a/rktl_launch/config/global_params.yaml
+++ b/rktl_launch/config/global_params.yaml
@@ -14,10 +14,10 @@ ball:
 
 # car dimensions and physical constants
 cars:
-  length: 0.12  # wheel center to center
+  length: 0.12  # front to rear wheel center to center, meters
   steering:
-    max_throw: 0.1826
-    rate: 0.9128
+    max_throw: 0.1826  # center to side, radians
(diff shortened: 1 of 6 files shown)


rktl_launch/config/global_params.yaml
```diff
@@ -14,11 +14,10 @@ ball:
 
 # car dimensions and physical constants
 cars:
-  width: 0.0762 # (3 in)
-  length: 0.2032 # (8 in)
+  length: 0.12  # wheel center to center
   steering:
-    max_throw: 0.26179938 # 15 degrees
-    rate: 0.52359877 # 30 degrees/s
+    max_throw: 0.1826
```
hmccarty, line 19: We should probably list units somewhere

hmccarty (changes requested):

rktl_launch/config/global_params.yaml
jrbaxter0: At the top of the file: `# ALL UNITS IN SI`. I can add them in though.

hmccarty: Oh true, didn't see that

hmccarty (approved):

PR #142, opened 2022-04-24
Xbox controller improvements

ajernandes:
I tested this in sim and it worked. 

Refactors the xbox_controller node that was originally written in like 10 minutes while I was waiting for thanksgiving dinner to be ready

Adds a boosting mechanic

Adds possibility for delay to understand the difficulty the autonomy experiences


jrbaxter0 (approved):
Style looks good. If it works, merge it.

PR #145, opened 2022-04-25
Patrol improve

jrbaxter0:
Some possible improvements to the patrol planner. Seem pretty good in sim but need to be tested IRL.


jrbaxter0:
We tested this at the meeting and it worked nicely so I'm merging it

PR #147, opened 2022-06-13
rktl Comments and Documentation

Abuynits:
Added comments for the parameters, return, and overview of each function in the following locations:
rktl_sim/src/simulator/sim.py
rktl_sim/nodes/simulator
rktl_autonomy/nodes/rocket_league_agent
rktl_autonomy/src/rktl_autonomy/_ros_interface.py
rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
rktl_autonomy/scripts/train_rocket_league.py

Note: I marked questions/things that I did not know with "TODO:"+ some text.

### DIFF
--- a/rktl_autonomy/nodes/rocket_league_agent
+++ b/rktl_autonomy/nodes/rocket_league_agent
@@ -10,7 +10,17 @@ from rktl_autonomy import RocketLeagueInterface
 from stable_baselines3 import PPO
 from os.path import expanduser
 import rospy
-
+"""
+overview:
+Load the weights for the specify model (PPO in this case)
+Create instance of the rocket_league_interface
+Reset the model for the rocket_league_interface
+Crete a PPO model based off the weights
+Infinite loop: 
+- predict the action based of the model
+- get the observation
+- step the action for the specific step based on the action 
+"""
 # create interface (and init ROS)
 env = RocketLeagueInterface(eval=True)
 
--- a/rktl_autonomy/scripts/train_rocket_league.py
+++ b/rktl_autonomy/scripts/train_rocket_league.py
@@ -15,13 +15,23 @@ from stable_baselines3.common.callbacks import CheckpointCallback
 from os.path import expanduser
 import uuid
 
-if __name__ == '__main__':      # this is required due to forking processes
+"""
+overview:
+All run env share a run_id
+This is where you specify launch argos: add to the env_kwargs
+Set the model that you will do to use for the training
+Log the training process for the specific run id
+Specify the stems and let the model learn wit these specific steps
+useful resource for stable_baselines3: https://stable-baselines3.readthedocs.io/_/downloads/en/master/pdf/
+"""
+
+if __name__ == '__main__':  # this is required due to forking processes
     run_id = str(uuid.uuid4())  # ALL running environments must share this
     print(f"RUN ID: {run_id}")
 
     # to pass launch args, add to env_kwargs: 'launch_args': ['render:=false', 'plot_log:=true']
-    env = make_vec_env(RocketLeagueInterface, env_kwargs={'run_id':run_id},
-            n_envs=24, vec_env_cls=SubprocVecEnv)
+    env = make_vec_env(RocketLeagueInterface, env_kwargs={'run_id': run_id},
+                       n_envs=24, vec_env_cls=SubprocVecEnv)
 
     model = PPO("MlpPolicy", env)
 
@@ -31,16 +41,16 @@ if __name__ == '__main__':      # this is required due to forking processes
     model.set_logger(logger)
 
     # log model weights
-    freq = 20833 # save 20 times
+    freq = 20833  # save 20 times
     # freq = steps / (n_saves * n_envs)
     callback = CheckpointCallback(save_freq=freq, save_path=log_dir)
 
     # run training
-    steps = 240000000 # 240M (10M sequential)
+    steps = 240000000  # 240M (10M sequential)
     print(f"training on {steps} steps")
     model.learn(total_timesteps=steps, callback=callback)
 
     # save final weights
     print("done training")
     model.save(log_dir + "/final_weights")
-    env.close() # this must be done to clean up other processes
+    env.close()  # this must be done to clean up other processes
--- a/rktl_autonomy/src/rktl_autonomy/_ros_interface.py
+++ b/rktl_autonomy/src/rktl_autonomy/_ros_interface.py
@@ -15,9 +15,11 @@ import rospy, roslaunch
 from rosgraph_msgs.msg import Clock
 from diagnostic_msgs.msg import DiagnosticStatus, KeyValue
 
+
 class SimTimeException(Exception):
     """For when advancing sim time does not go as planned."""
     pass
+
 
 class ROSInterface(Env):
     """Extension of the Gym environment class for all specific interfaces
@@ -47,6 +49,19 @@ class ROSInterface(Env):
             launch_file: if training, launch file to be used (ex: ['rktl_autonomy', 'rocket_league_train.launch'])
             launch_args: if training, arguments to be passed to roslaunch (ex: ['render:=true', rate:=10])
             run_id: if training, used to prevent deadlocks. if logging, run_id describes where to save files. Default is randomly generated
+        Overview:
+        If eval = false (training agnet):
+            - training- try to open launch files
+            - generate uuid (our use provided )
+            - roslauch setup and configuration for the specific environment
+            - Init the time space for the node
+            -
+        If eval = true: (in existing ROS env) (not training)
+            - use existing ROS network, and just create a node
+
+        For all:
+        create a log publisher
+        setup generic variables for login
         """
         super().__init__()
         self.__EVAL_MODE = eval
@@ -76,10 +91,11 @@ class ROSInterface(Env):
             ros_id = roslaunch.rlutil.get_or_generate_uuid(None, False)
             roslaunch.configure_logging(ros_id)
             launch_file = roslaunch.rlutil.resolve_launch_arguments(launch_file)[0]
-            launch_args = [f'render:={port==11311}', f'plot_log:={port==11311}'] + launch_args + [f'agent_name:={node_name}']
+            launch_args = [f'render:={port == 11311}', f'plot_log:={port == 11311}'] + launch_args + [
+                f'agent_name:={node_name}']
             launch = roslaunch.parent.ROSLaunchParent(ros_id, [(launch_file, launch_args)], port=port)
             launch.start()
-            self.close = lambda : launch.shutdown()
+            self.close = lambda: launch.shutdown()
             # initialize self
             os.environ['ROS_MASTER_URI'] = f'http://localhost:{port}'
             rospy.init_node(node_name)
@@ -91,6 +107,7 @@ class ROSInterface(Env):
 
         # private variables
         self._cond = Condition()
+        # TODO: why cant merge to if statement above?
 
         # additional set up for training
         if not self.__EVAL_MODE:
@@ -114,6 +131,7 @@ class ROSInterface(Env):
 
     def step(self, action):
         """
+        high-level:
         Implementation of gym.Env.step. This function will intentionally block
         if the ROS environment is not ready.
 
@@ -128,16 +146,22 @@ class ROSInterface(Env):
             reward (float) : amount of reward returned after previous action
             done (bool): whether the episode has ended, in which case further step() calls will return undefined results
             info (dict): contains auxiliary diagnostic information (helpful for debugging, and sometimes learning)
+
+        overview:
+        Runs: _clear_state, _publish_action, _step_time_and_wait_for_state (all abstract methods that are implemented by the child)
+        Gets the simulation state via: _get_state
         """
         self._clear_state()
         self._publish_action(action)
         self.__step_time_and_wait_for_state()
         state = self._get_state()
-        self.__net_reward += state[1]   # logging
+        self.__net_reward += state[1]  # logging
         return state
 
     def reset(self):
-        """Resets the environment to an initial state and returns an initial observation.
+        """
+        High-level:
+        Resets the environment to an initial state and returns an initial observation.
 
         Note that this function should not reset the environment's random
         number generator(s); random variables in the environment's state should
@@ -146,13 +170,27 @@ class ROSInterface(Env):
         a new episode, independent of previous episodes.
         Returns:
             observation (object): the initial observation.
+        Overview:
+        Checks if have a new state ready via: _has_state
+        If so: gathers information: episode #, net reward, duration of the episode
+        Update the message log with these parameters by publishing it
+        increment the episode count
+        reset the reward accumulated for the past episode
+
+        reset the sim:
+        if eval = false ( training the agent)
+        - reset the environment with overidden function (abstract method)
+        for all:
+        - reset the ROS interface (abstract method)
+        -
         """
+
         if self._has_state():
             # generate log
             info = {
-                'episode'    : self.__episode,
-                'net_reward' : self.__net_reward,
-                'duration'   : (rospy.Time.now() - self.__start_time).to_sec()
+                'episode': self.__episode,
+                'net_reward': self.__net_reward,
+                'duration': (rospy.Time.now() - self.__start_time).to_sec()
             }
             info.update(self._get_state()[3])
             # send message
@@ -172,8 +210,17 @@ class ROSInterface(Env):
             self._reset_env()
         self._reset_self()
         self.__step_time_and_wait_for_state(5)
-        self.__start_time = rospy.Time.now()    # logging
+        self.__start_time = rospy.Time.now()  # logging
         return self._get_state()[0]
+
+    """
+    input: 
+    - number of retries: step time until state is known
+    overview:
+    Increment time and clock, take input of number of tries and do them until figure stuff out
+    Call __wait_once_for_state for the provided number of retries
+    increment the time if retrying wait_once_for_state
+    """
 
     def __step_time_and_wait_for_state(self, max_retries=1):
         """Step time until a state is known."""
@@ -191,10 +238,11 @@ class ROSInterface(Env):
                     retries += 1
         else:
             while not self.__wait_once_for_state():
-                pass    # idle wait
+                pass  # idle wait
 
+    """overview: Wait and allow other threads to run."""
     def __wait_once_for_state(self):
-        """Wait and allow other threads to run."""
+
         with self._cond:
             has_state = self._cond.wait_for(self._has_state, 0.25)
         if rospy.is_shutdown():
@@ -202,44 +250,49 @@ class ROSInterface(Env):
         return has_state
 
     # All the below abstract methods / properties must be implemented by subclasses
+    """overview: The Space object corresponding to valid actions."""
     @property
     @abstractmethod
     def action_space(self):
-        """The Space object corresponding to valid actions."""
+
         raise NotImplementedError
 
+    """Overview: The Space object corresponding to valid observations."""
     @property
     @abstractmethod
     def observation_space(self):
-        """The Space object corresponding to valid observations."""
+
         raise NotImplementedError
 
+    """Overview: Reset environment for a new episode."""
     @abstractmethod
     def _reset_env(self):
-        """Reset environment for a new episode."""
+
         raise NotImplementedError
 
+    """Overview: Reset internally for a new episode."""
     @abstractmethod
     def _reset_self(self):
-        """Reset internally for a new episode."""
+
         raise NotImplementedError
 
+    """Overview: Determine if the new state is ready."""
     @abstractmethod
     def _has_state(self):
-        """Determine if the new state is ready."""
+
         raise NotImplementedError
 
+    """Overview: Clear state variables / flags in preparation for new ones."""
     @abstractmethod
     def _clear_state(self):
-        """Clear state variables / flags in preparation for new ones."""
         raise NotImplementedError
 
+    """Overview: Get state tuple (observation, reward, done, info)."""
     @abstractmethod
     def _get_state(self):
-        """Get state tuple (observation, reward, done, info)."""
         raise NotImplementedError
 
+    """Overview: Publish an action to the ROS network."""
     @abstractmethod
     def _publish_action(self, action):
-        """Publish an action to the ROS network."""
         raise NotImplementedError
--- a/rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
+++ b/rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
@@ -21,9 +21,12 @@ from tf.transformations import euler_from_quaternion
 from enum import IntEnum, unique, auto
 from math import pi, tan
 
+
 @unique
 class CarActions(IntEnum):
-    """Possible actions for car."""
+    """high level: Possible actions for car.
+    currently using discrete action space
+    """
     STOP = 0
     FWD_LEFT = auto()
     FWD_RIGHT = auto()
@@ -33,17 +36,57 @@ class CarActions(IntEnum):
     REV = auto()
     SIZE = auto()
 
+
 class RocketLeagueInterface(ROSInterface):
-    """ROS interface for the Rocket League."""
-    def __init__(self, eval=False, launch_file=['rktl_autonomy', 'rocket_league_train.launch'], launch_args=[], run_id=None):
-        super().__init__(node_name='rocket_league_agent', eval=eval, launch_file=launch_file, launch_args=launch_args, run_id=run_id)
+    """ROS interface for the Rocket League.
+    param:
+    eval: specify whether ytou want to train data or not (look at _ros_interface.py init function)
+    launch_file: used in _ros_interface.py to configure the training for roslaucnh(look at _ros_interface.py init function)
+    launch args: specify the files to be used in launching (look at _ros_interface.py init function)
+    run_id: the id for this specific training run (look at _ros_interface.py init function)
+    Overview:
+        Extends ROS interface: init the super
+        Set some car constants
+        - min/max vel
+        - curvature
+        Override the car constants if you find that the parameters from the action space are < for the min and if they are > for the max of the car’s action space
+        Set the parameters for the observed objects on the field:
+        - goal depth
+        - field dimensions
+        - max velocity (angular and linear)
+
+        set the parameters for the learning part of RL:
+        TODO: all are currently constant: want to add functionality to randomize maybe?
+        - max time
+        - constant reward
+        - ball distance reward
+        - goal distance reward
+        - loss reward
+        - win reward
+        - reverse reward
+        - hitting the wall reward
+
+        Define a reset function for this specific environment: returned from rospy
+        publish:
+        - car0 command
+        - sim_reset
+        subscribe:
+        - car0/odometry
+        - ball/odometry
+        - match status
+    """
+
+    def __init__(self, eval=False, launch_file=['rktl_autonomy', 'rocket_league_train.launch'], launch_args=[],
+                 run_id=None):
+        super().__init__(node_name='rocket_league_agent', eval=eval, launch_file=launch_file, launch_args=launch_args,
+                         run_id=run_id)
 
         ## Constants
         # Actions
         self._MIN_VELOCITY = -rospy.get_param('/cars/throttle/max_speed')
-        self._MAX_VELOCITY =  rospy.get_param('/cars/throttle/max_speed')
+        self._MAX_VELOCITY = rospy.get_param('/cars/throttle/max_speed')
         self._MIN_CURVATURE = -tan(rospy.get_param('/cars/steering/max_throw')) / rospy.get_param('cars/length')
-        self._MAX_CURVATURE =  tan(rospy.get_param('/cars/steering/max_throw')) / rospy.get_param('cars/length')
+        self._MAX_CURVATURE = tan(rospy.get_param('/cars/steering/max_throw')) / rospy.get_param('cars/length')
 
         # Action space overrides
         if rospy.has_param('~action_space/velocity/min'):
@@ -68,7 +111,7 @@ class RocketLeagueInterface(ROSInterface):
         self._FIELD_LENGTH = rospy.get_param('/field/length')
         self._GOAL_DEPTH = rospy.get_param('~observation/goal_depth', 0.075)
         self._MAX_OBS_VEL = rospy.get_param('~observation/velocity/max_abs', 3.0)
-        self._MAX_OBS_ANG_VEL = rospy.get_param('~observation/angular_velocity/max_abs', 2*pi)
+        self._MAX_OBS_ANG_VEL = rospy.get_param('~observation/angular_velocity/max_abs', 2 * pi)
 
         # Learning
         self._MAX_TIME = rospy.get_param('~max_episode_time', 30.0)
@@ -100,36 +143,49 @@ class RocketLeagueInterface(ROSInterface):
         if not eval:
             rospy.wait_for_service('sim_reset')
 
+    """
+    high level: The Space object corresponding to valid actions.
+    """
+
     @property
     def action_space(self):
-        """The Space object corresponding to valid actions."""
         return Discrete(CarActions.SIZE)
+
+    """
+    high level: The Space object corresponding to valid observations.
+    returns the lower and upper bound of the field
+    TODO: why upper and lower bounds?    
+    """
 
     @property
     def observation_space(self):
-        """The Space object corresponding to valid observations."""
         return Box(
             # x, y, theta, v, omega (car)
             # x, y, vx, vy (ball)
             low=np.array([
-                -(self._FIELD_LENGTH/2) - self._GOAL_DEPTH,
-                -self._FIELD_WIDTH/2, -pi,
+                -(self._FIELD_LENGTH / 2) - self._GOAL_DEPTH,
+                -self._FIELD_WIDTH / 2, -pi,
                 -self._MAX_OBS_VEL, -self._MAX_OBS_ANG_VEL,
-                -(self._FIELD_LENGTH/2) - self._GOAL_DEPTH,
-                -self._FIELD_WIDTH/2,
+                -(self._FIELD_LENGTH / 2) - self._GOAL_DEPTH,
+                -self._FIELD_WIDTH / 2,
                 -self._MAX_OBS_VEL, -self._MAX_OBS_VEL],
                 dtype=np.float32),
             high=np.array([
-                (self._FIELD_LENGTH/2) + self._GOAL_DEPTH,
-                self._FIELD_WIDTH/2, pi,
+                (self._FIELD_LENGTH / 2) + self._GOAL_DEPTH,
+                self._FIELD_WIDTH / 2, pi,
                 self._MAX_OBS_VEL, self._MAX_OBS_ANG_VEL,
-                (self._FIELD_LENGTH/2) + self._GOAL_DEPTH,
-                self._FIELD_WIDTH/2,
+                (self._FIELD_LENGTH / 2) + self._GOAL_DEPTH,
+                self._FIELD_WIDTH / 2,
                 self._MAX_OBS_VEL, self._MAX_OBS_VEL],
                 dtype=np.float32))
-#TODO: this is the function that is called by the reset to the sim from the rocket_league_interface
+
+    # this is the function that is called by the reset to the sim from the rocket_league_interface
     def _reset_env(self):
-        """Reset environment for a new training episode."""
+        """
+             high level: Reset environment for a new training episode.
+             Reset the environment for the car
+             #TODO: important: call subscriber function
+        """
         self._reset_srv.call()
 
     def _reset_self(self):
@@ -140,9 +196,9 @@ class RocketLeagueInterface(ROSInterface):
     def _has_state(self):
         """Determine if the new state is ready."""
         return (
-            self._car_odom is not None and
-            self._ball_odom is not None and
-            self._score is not None)
+                self._car_odom is not None and
+                self._ball_odom is not None and
+                self._score is not None)
 
     def _clear_state(self):
         """Clear state variables / flags in preparation for new ones."""
@@ -151,7 +207,21 @@ class RocketLeagueInterface(ROSInterface):
         self._score = None
 
     def _get_state(self):
-        """Get state tuple (observation, reward, done, info)."""
+        """
+        Get state tuple (observation, reward, done, info).
+        Overview:
+            Observation:
+                Check if the observation is within the limits: make sure that the car is in bounds
+                If out of blinds: clip the observation space; by the lwo and high, and then continue processing
+            Time:
+                Set the current start time, checn whether the current time, self time is greater tha maximum allocation time:
+            Determine the reward:
+                Start with the constant reward, add , the distance traveled by the ball( ball distance reward
+                Get the change in distance to the goal, multiply by the reward to the distance
+                Calculate the score: if win, add 100, add lose score
+                Wall reward: check if hit the wall, if not add to it, if yet, remove from it
+
+        """
         assert self._has_state()
 
         # combine car / ball odoms for observation
@@ -179,7 +249,7 @@ class RocketLeagueInterface(ROSInterface):
         ball_dist_sq = np.sum(np.square(ball[0:2] - car[0:2]))
         reward += self._BALL_DISTANCE_REWARD * ball_dist_sq
 
-        goal_dist_sq = np.sum(np.square(ball[0:2] - np.array([self._FIELD_LENGTH/2, 0])))
+        goal_dist_sq = np.sum(np.square(ball[0:2] - np.array([self._FIELD_LENGTH / 2, 0])))
         reward += self._GOAL_DISTANCE_REWARD * goal_dist_sq
 
         if self._score != 0:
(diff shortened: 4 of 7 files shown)


rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
...
+                # bc not have max and min, this means that we have more dicts and vals to loop over
+                for key in rospy_param:
+                    #recurse down the path, using the key as the continuation
+                    rospy_param[key] = self.get_sim_param(path + f'$/{key}')
+                return rospy_param
+
+            elif type_rospy == float or type_rospy == int or type_rospy == list:
+                return rospy.get_param(path)
+
+        raise self.InvalidRospyParam(path)
```
hmccarty, line 519: We don't want to raise an exception in a ROS node. Instead we should handle error, log and shutdown cleanly.

rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
...
+def delete_all_cars(self):
+    car_configs = rospy.get_param('~cars', [])
+    self.reset_lock.acquire()
+    for car_config in car_configs:
+
+        if 'name' not in car_config:
+            rospy.signal_shutdown('no "name" set for car config in sim')
+        car_name = car_config['name']
+        if 'randomize_pose' in car_config:
+            init_pose = None
```
hmccarty, line 580: I think this is leftover from previous implementation

rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
...
+    return DeleteCarResponse(True)
+
+
+def generate_all_cars(self):
+    car_configs = rospy.get_param('~cars', [])
+
+    for car_config in car_configs:
+        init_pose = None
+        if 'init_pose' in car_config:
+            init_pose = car_config['init_pose']
```
hmccarty, line 615: We should be using `get_sim_param` for this

rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
...
+    """
+    sensor_noise = rospy.get_sim_param('~sensor_noise', None)
+    ball_noise = None
+    car_noise = None
+    if sensor_noise is not None and self.mode == SimulatorMode.REALISTIC:
+        ball_noise = sensor_noise.get('ball', None)
+        car_noise = sensor_noise.get('car', None)
+    self.car_noise = car_noise
+    self.ball_noise = ball_noise
+    return car_noise, ball_noise
```
hmccarty, line 660: We should either set member variables or return the variable, doing both is a little overkill

rktl_sim/nodes/simulator
```diff
@@ -68,75 +122,22 @@ class Simulator(object):
...
 
         # Setting sensor noise
-        sensor_noise = rospy.get_param('~sensor_noise', None)
-        ball_noise = None
-        car_noise = None
-        if sensor_noise is not None and self.mode == SimulatorMode.REALISTIC:
-            ball_noise = sensor_noise.get('ball', None)
-            car_noise = sensor_noise.get('car', None)
-        self.car_noise = car_noise
+        self.car_noise, self.ball_noise = self.set_sensor_noise()
```
hmccarty, line 125: I think this is a case that probably doesn't require a function breakout

rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
...
+            del self.car_pose_pubs[car_name]
+    # reseting vars to be safe
+    self.car_ids = {}
+    self.car_pose_pubs = {}
+    self.car_odom_pubs = {}
+    self.car_effort_subs = {}
+    self.car_cmd_subs = {}
+    self.reset_lock.release()
+
+    return DeleteCarResponse(True)
```
hmccarty, line 606: To be used, this code needs to be in the service callback. Applied for the generate function too.

rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
...
+                                        PoseWithCovarianceStamped, queue_size=1)
+        ball_odom_pub = rospy.Publisher('/ball/odom_truth',
+                                        Odometry, queue_size=1)
+        return ball_pose_pub, ball_odom_pub
+    elif self.mode == SimulatorMode.IDEAL:
+        ball_odom_pub = rospy.Publisher('/ball/odom',
+                                        Odometry, queue_size=1)
+        return None, ball_odom_pub
+
+    return None, None
```
hmccarty, line 687: I think this is also tame enough that we can leave it in the original init function. Function breakout should be reserved for cases where we want to hide the detail of modifying/mutating data. If we are just setting or declaring variables, then those operations should happen in the init function.

rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
...
+    # if too constrained in car length between what we detect and what it is - might learn differently
+    # use the get_sim_param in each function case: makes the most sense
+    car_properties = {'length': rospy.get_sim_param('/cars/length'),
+                      'max_speed': self.get_sim_param("/cars/throttle/max_speed"),
+                      'steering_throw': self.get_sim_param("/cars/throttle/max_steering_throw"),
+                      'throttle_tau': rospy.get_sim_param("/cars/throttle/tau"),
+                      'steering_rate': rospy.get_sim_param("/cars/steering/rate"),
+                      'simulate_effort': (self.mode == SimulatorMode.REALISTIC)}
+    self.car_properties = car_properties
+    return car_properties
```
hmccarty, line 568: See note on `create_ball`, I think same philosophy applies here. Over the phone we talked about two parameter modalities: init and instantiate. Instantiate will be updated by autonomy on each reset and init configured once on startup. I might've originally suggested that car_props are an instantiate example, but rethinking it, they should be init params. Meaning we should only set them in the init function.

hmccarty (changes requested):
Left a lot of comments. Feel free to review, then call on discord if you have questions / disagree (which is encouraged).

For next time, we should try to move a lot of those comments to personal documentation.

rktl_sim/nodes/simulator
```diff
@@ -358,5 +449,244 @@ class Simulator(object):
         self.last_time = now
         self.reset_lock.release()
 
+    def get_sim_dict(self, path):
+
+        if not rospy.get_param(path):
+            raise self.InvalidRospyParam(path)
```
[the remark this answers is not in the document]
Abuynits: the idea of raising the exception is to also print out the file path and exit the code cleanly. If you go to the IvalidRospyParam exception, you will see that it prints the exception and exits the simulation cleanly via rospy.logfatal(). Would this work, or should I replace all of my exceptions with rospy.logfatal90? I just thought that it would be nicer to print the file path that causes the error before exiting the simulation.

rktl_autonomy/nodes/rocket_league_agent
```diff
@@ -10,7 +10,17 @@ from rktl_autonomy import RocketLeagueInterface
...
+overview:
+Load the weights for the specify model (PPO in this case)
+Create instance of the rocket_league_interface
+Reset the model for the rocket_league_interface
+Crete a PPO model based off the weights
+Infinite loop: 
+- predict the action based of the model
+- get the observation
+- step the action for the specific step based on the action 
+"""
```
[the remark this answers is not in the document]
Abuynits: Personally, I think that by providing these notes, a person will be able to start working on the project faster (getting caught up and understanding the repo better). They have helped me out a lot (probably bc they are personal notes lol). 
Is it possible to remove them from the final commit and then add them in? 
I can also write more straightforward documentation as in the one that you recommend on lines 1-7 or those present in other functions that were not written by me. I just really think that there should be a blurb of comments to describe key functionality in the code and for each function.

rktl_autonomy/scripts/train_rocket_league.py
```diff
@@ -31,16 +41,16 @@
...
-    steps = 240000000 # 240M (10M sequential)
+    steps = 240000000  # 240M (10M sequential)
     print(f"training on {steps} steps")
     model.learn(total_timesteps=steps, callback=callback)
 
     # save final weights
     print("done training")
     model.save(log_dir + "/final_weights")
-    env.close() # this must be done to clean up other processes
+    env.close()  # this must be done to clean up other processes
```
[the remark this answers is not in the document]
Abuynits: done!

rktl_sim/src/simulator/car.py
```diff
@@ -113,14 +120,42 @@ def get_velocity(self):
...
+
+        # system state
         self._v_rear = 0.0
         self._psi = 0.0
 
+        # model configuration
+        p.resetBasePositionAndOrientation(
+            self.id, [0., 0., pos[2]], p.getQuaternionFromEuler([0., 0., 0.]))
+
+        self.joint_ids = (1, 0, 2)  # X, Y, W
```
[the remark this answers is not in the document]
Abuynits: done!

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -191,55 +238,61 @@ def __step_time_and_wait_for_state(self, max_retries=1):
...
-        """Reset environment for a new episode."""
+
         raise NotImplementedError
 
+    """Overview: Reset internally for a new episode."""
     @abstractmethod
     def _reset_self(self):
-        """Reset internally for a new episode."""
+
         raise NotImplementedError
```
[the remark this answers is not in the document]
Abuynits: done! 
 moved all docstrings to inside the functions for all areas where I added changes
 removed overview in all areas  
 added variables that were not declared in function initialization'

rktl_sim/urdf/walls.urdf
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" ?>
 <!-- =================================================================================== -->
-<!-- |    This document was autogenerated by xacro from /home/hsan/catkin_ws/src/rktl_3/rktl_sim/urdf/walls.urdf.xacro | -->
+<!-- |    This document was autogenerated by xacro from /home/abuynits/catkin_ws/src/rocket_league/rktl_sim/urdf/walls.urdf.xacro | -->
```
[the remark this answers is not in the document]
Abuynits: I added the whole folder to the .git ignore ie rktl_sim/urdf. All of the files in it are autogenerated. is this what you are looking for?

rktl_sim/src/simulator/car.py
```diff
@@ -113,14 +120,42 @@ def get_velocity(self):
...
+
+    def check_overlap(self, pos):
+        """
+        returns whether the two positions of the car will overlap
+        @param pos: the position of the other car
+        @return: bolean if they overlap (true = overlap)
+        """
+        dist = math.sqrt(
+            (pos[0] - self.pos[0]) * (pos[0] - self.pos[0]) + (pos[0] - self.pos[1]) * (pos[1] - self.pos[1]))
+        return dist < self._LENGTH
```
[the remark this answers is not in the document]
Abuynits: Out of curiosity what objects would we have that would be larger than a car? 
If so, it would make sense to create a general object class with general methods such as reset, step, get velocity, and so on and have the car, ball and other objects extend it for simplicity. That way I can have a general check_overlap function in that clas that would check the distance from each object using the largest length of the car.

(I changed the property from self.pos to self.init_pos

rktl_sim/nodes/simulator
Abuynits: I made the InvalidRospyParam exception to be in charge of the error. the implementation is below for this specific exception: 
`   class InvalidRospyParam(Exception):
"""exception when using invalid arguement"""
print(Exception)
rospy.logfatal()
`
this is located in the simulator node

Abuynits: Does this work, or are you looking over for something else?

rktl_sim/nodes/simulator
Abuynits: the delete_all_cars function loops over all of the cars and deletes them. while the delete_car_cb only deletes 1 car. I have attached a screenshot comparing the two segments of code:
image
In addition, the delete_all_cars function remains unused as it was part of my initial solution of deleting all cars, and then regenerating them from scratch. Now, I just loop over the already generated cars and create them. I just left it in bc it might be useful in the future. 

The same applies for create_car_cb (creates only 1 car in a callback), and generate_all_cars (loops over all cars and generates them)
image

I only use generate_all_cars in the initialization of the simulator class. I don't use any of the callbacks in my changes, but I am sure that they are used in other places. Generate_all_cars is called only once. What should I do? I can add delete_all_cars_cb and generate_all_cb, as new call back functions, and I will loop over all of cars and delete them. I can also modify the current create_car_cb and delete_car_cb to delete/create all cars.

also out of curiosity, what is the req parameter in the create_car_cb and the delete_car_cb. Is it the path to the specific car?

sorry for the long message!

rktl_sim/nodes/simulator
Abuynits: I removed it!

rktl_sim/nodes/simulator
Abuynits: done!

rktl_sim/src/simulator/car.py
```diff
@@ -10,37 +10,44 @@
...
 
         # physical constants
+        self.set_car_properties(car_properties)
+
+        # urdf configuration
+        self.body_link_id = 1
+
+        self.reset(pos, orient)
+
+    def set_car_properties(self, car_properties):
```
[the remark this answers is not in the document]
Abuynits: ye that makes sense.

rktl_sim/nodes/simulator
Abuynits: done! I also updated the get_sim_param and get_sim_dict function with another variable that specifies if you are fine with receiving None as a return or having an error thrown.

rktl_sim/nodes/simulator
Abuynits: Cool!

rktl_sim/src/simulator/sim.py
```diff
@@ -235,16 +353,90 @@ def get_ball_pose(self, add_noise=False):
...
+    - try to get the ini_pos and init_orientation if it was provided
+    - If the orientation and position were not initialized, then randomize them, and reset the car
+    """
+
+    def reset(self, spawn_bounds, car_properties):
         self.scored = False
         self.winner = None
         self.touched_last = None
+        self.spawn_bounds = spawn_bounds
+        self.reset_ball()
```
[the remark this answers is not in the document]
Abuynits: I will leave the functionality of reseting the car properties if we would want to do that in the future. The reset of car properties will still occur however, because the car properties themselves wont change, nothing will change here. I think that its a good feature to leave in if we would want to play around with handling the rest of car properties on the simulator side.

rktl_autonomy/nodes/rocket_league_agent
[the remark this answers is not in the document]
hmccarty: Wiki would be a good place for onboarding notes.

For code comments, you would want to split these paragraphs into: file summaries (top of file), function summaries (top of function), deobfuscation comments (inline). You want to avoid stating the obvious, for sake of clarity, especially since python is **usually** very readable. Sometimes you may opt for rewriting code, as opposed to adding comments.

rktl_sim/urdf/walls.urdf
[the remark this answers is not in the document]
hmccarty: No, `xacro` files generate the `urdf` files. So we want to leave those out of the `.gitignore`. Since we want to keep some `urdf` files, just explicitly state the autogenerated ones.

rktl_sim/nodes/simulator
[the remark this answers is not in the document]
hmccarty: Printing and logging serve different functions. Using the `print` function will hide errors when stdout is ignored by ros and doesn't store for later reference. The same info can be printed and stored using `logfatal` (see other prior examples) without requiring a need for an exception breakout.

rktl_sim/nodes/simulator
hmccarty: If we aren't going to use the function, then I would recommend we delete it to keep things slim. Since deletion of each car only requires pulling out the list of IDs, I think we only need the `delete_car_cb` / `create_car_cb`. It feels good to keep things concise.

The `req` param is the service request message defined in `rktl_msgs`. We should add mypy type checks at some point.

rktl_autonomy/nodes/rocket_league_agent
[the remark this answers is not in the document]
Abuynits: I reworked all of my documentation and code comments for the repo. Can you take another look?

rktl_sim/urdf/walls.urdf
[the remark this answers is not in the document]
Abuynits: Cool!
These are the ones I added that were autogenerated
rktl_sim/urdf/walls.urdf
rktl_sim/urdf/goal_a.urdf
rktl_sim/urdf/goal_b.urdf
rktl_sim/urdf/walls.urdf

rktl_sim/nodes/simulator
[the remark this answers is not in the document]
Abuynits: done!

rktl_sim/nodes/simulator
Abuynits: Im down to rework it this way. The only problem that I cannot work around is that the create/delete_car_cb uses are req which has things such as init_pose, init_yaw, and I'm certain that it has more attributes. My current function of generate_all_cars does all of this via looking over the car_configs list and doing the same functionality (ie for new init_positions, it just reads from rospy again), while most of it remain the same. (The same applies to the delete_car_cb, but I do agree that if we don't use it, I should remove it.

Therefore, my question is this: how do I rework create_car_cb so that I can loop over all of the cars in the generate_all_cars function without having a req but rather having the cars id if it has certain functionality that relies on the data held in a req?

Would it help to call bc I feel that I might be getting confused lol.

rktl_autonomy/scripts/train_rocket_league.py
```diff
@@ -15,13 +15,22 @@
...
+"""
+All run env share a run_id
+This is where you specify launch argos: add to the env_kwargs
+Set the model that you will do to use for the training
+Log the training process for the specific run id
+Specify the stems and let the model learn wit these specific steps
+useful resource for stable_baselines3: https://stable-baselines3.readthedocs.io/_/downloads/en/master/pdf/
+"""
+
+if __name__ == '__main__':  # this is required due to forking processes
```
hmccarty, line 27: I still find this comment pretty cryptic and not sure if it adds value to this file. We can add the stable baselines link to the wiki.

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -163,83 +169,95 @@ def reset(self):
             msg.hardware_id = self.__LOG_ID
             msg.values = [KeyValue(key=key, value=str(value)) for key, value in info.items()]
             self.__log_pub.publish(msg)
-            # update variables (update time after reset)
+            # increment the episode count
```
hmccarty, line 172: This is one of those comments that likely states the obvious, we can probably go without a comment here

rktl_sim/src/simulator/car.py
```diff
@@ -104,23 +122,53 @@ def get_pose(self, noise=None):
...
                           [math.sin(heading), math.cos(heading), 0.],
-                         [0., 0., 1.]], dtype=np.float)
+                          [0., 0., 1.]], dtype=np.float)
         linear = r_inv @ linear
-        return (linear, angular)
+        return linear, angular
 
     def reset(self, pos, orient):
+        """
+        uses the position and the orientation to reset system state and car configuration
```
hmccarty, line 143: In other comments, we use capitalization and punctuation, we should strive for consistency

rktl_sim/nodes/simulator
hmccarty: That's partially my fault with changing my thoughts on the design. I'll try to layout what I now think is the best approach below:

We have two types of parameters: configuration-based and instance-based.

Configuration-based parameters determine the kinematics and dynamics of the sim. A few examples: car size, sim timestep, friction, etc.

Instance-based parameters determine the initial state of a new run. A few examples: car pose, ball pose, etc.

For any sim param, you should have the option of either specifiying a value:
```
car_length: X
``` 
Or a range:
```
car_length:
  max: X
  min: Y
```
If given a range, the sim will automatically select a value on each instance it retrieves its value.

On startup, we load both configuration and instance parameters and create an initial setup.

Throughout usage of the sim, users and other processes will want to create new instances. Previously, we used a service to do so, since it's human readable and understandable. This is less than ideal for the autonomy, since it may want to modify the instance value of each car, then reset to a new instance.

So I'm thinking our approach is this:

One global reset service. This starts a new clean slate (deletes old objects first), then reloads configuration params using `get_sim_param`, and sets up cars and other objects using the instance params (again using `get_sim_param`). This is perfect for other processes, since they can easily modify rospy params, then call the global reset. But now we've made it difficult for users who want to control instance params.

To solve this, we use the existing create / destroy car services as wrappers for param modification. So instead of the create car service actualing creating a new car in the sim object, it simply sets the instance param for us using `rospy.set_param()`.

How does this sound?

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
Abuynits: Ye, I should have caught that...

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -116,44 +120,46 @@ def step(self, action):
...
+        @return: observation a tuple of the following:
+            (object): agent's observation of the current environment
             reward (float) : amount of reward returned after previous action
             done (bool): whether the episode has ended, in which case further step() calls will return undefined results
             info (dict): contains auxiliary diagnostic information (helpful for debugging, and sometimes learning)
+
         """
+        # Run one timestep of the environment's dynamics. When end of
+        # episode is reached, you are responsible for calling `reset()`
+        # to reset this environment's state.
```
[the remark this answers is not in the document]
Abuynits: I moved it in!

rktl_sim/src/simulator/car.py
Abuynits: I worked through all of the files and added capitalization, punctuation, and fixed grammar along the way. I think that I covered all of the files, but I'm not sure. Lmk if there are some that I missed?

rktl_sim/src/simulator/sim.py
```diff
@@ -236,15 +311,78 @@ def get_ball_pose(self, add_noise=False):
         return pos, p.getQuaternionFromEuler([0, 0, 0])
 
     def get_ball_velocity(self):
+        """
+        returns the velocity of the ball
```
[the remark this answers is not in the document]
Abuynits: Cool, I removed it and similar comments from sim.py

rktl_sim/urdf/walls.urdf
```diff
@@ -1,6 +1,6 @@
...
 <!-- |    EDITING THIS FILE BY HAND IS NOT RECOMMENDED                                 | -->
 <!-- =================================================================================== -->
```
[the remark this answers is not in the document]
Abuynits: Cool, i removed it from being tracked by the repo!

rktl_autonomy/scripts/train_rocket_league.py
Abuynits: I think that it was my own notes that I wrote that I forgot to delete. I left the baseline 3 resource in there for now, at the top of the file comment, so that it will not be lost.

rktl_sim/nodes/simulator
Abuynits: That sounds like a great approach. I just want to clarify how a run of the simulator works to see if I am understanding it the same as you:

The sim is initialized, creating: 
configuration-based parameters ie car properties, time step, car turning radius etc via get_sim_param.
instance-based parameters ie car list, ball, field size etc
- the init function of the simulator creates configuration-based parameters while calling functions in the sim.py to create instance-based parameters
- uses the create_car_cb to create cars

In between runs, we keep the configuration-based properties constant, while we loop over the instance-based parameters and reset their values where we don't create and delete objects but rather reset their individual parameters (ie new spawn locations and new field size)
- the reset function handles the reset of instance parameters that are all stored mostly in the sim.py file.

at the end of the sim, we physically delete all of the instance-based parameters ( cars, ball objects), and all of the configuration-based parameters.
- with the delete_car_cb

However, I still don't understand how I use these functions with the req parameter if I do not have it at the init function and at the deletion step from the sim side. Would it be safe to keep my existing functions of delete_all_cars and create_all_cars as they loop over the cars present in the simulator through their IDs and not their reqs? Another solution might be to perform a type check on the delete_car_cb where I will pass an id into the req from the simulator side. a req is a dict while an id is not a dict lol. If the id is passed in then I will create a car off of it, without breaking any existing functionality that is caused by using a req for car creation. The same applies to car deletion.

Would this work, or am I still confused about the design?

Abuynits: Todays call really helped me understand how everything functions! I completely reworked the reinitialization of the sim such that whenever the create_car_cb / delete_car_cb is invoked, the instance-based parameters are reset (params like props, engine, ball init position, car properties, car init pos and velocity etc..) while the config-based parameters are kept constant (urdf, rate, frame id, sim mode, etc...). In addition, I did not actually have to create/delete all of the objects, but I loop over all of them and reset their parameters, saving time. I also added two more callback functions (delete/create_all_cars) just because I already had the functions written and used in other places in the sim. Let me know if there is anything else that I am missing!

rktl_sim/nodes/simulator
```diff
@@ -160,60 +122,70 @@ class Simulator(object):
...
+        """
+
         if param is None:
             rospy.signal_shutdown('no urdf path set for "{}"'.format(param))
 
         i = 0
         while not os.path.isfile(param) and i < 5:
-            rospy.sleep(0.1) # Wait for xacro build
+            # Wait for xacro build.
+            rospy.sleep(0.1)
```
hmccarty, line 136: I prefer the previous 1 line for conciseness

rktl_sim/nodes/simulator
```diff
@@ -160,60 +122,70 @@ class Simulator(object):
...
 
+        # Subscribers
         car_id = self.car_ids[req.name]
         self.car_effort_subs[req.name] = rospy.Subscriber(
             f'/cars/{req.name}/effort', ControlEffort,
             self.effort_cb, callback_args=car_id)
-        self.car_cmd_subs[req.name] =  rospy.Subscriber(
+        self.car_cmd_subs[req.name] = rospy.Subscriber(
             f'/cars/{req.name}/command', ControlCommand,
             self.cmd_cb, callback_args=car_id)
```
hmccarty, line 185: So we still want to change this function to do the set param implementation. `sim.create_car` should only be called in our 'global' reset function.

rktl_sim/src/simulator/car.py
```diff
@@ -69,31 +84,36 @@ def step(self, dt):
...
         else:
             body_vel = self.cmd[0]
             if abs(body_vel) > self._MAX_SPEED:
                 body_vel = math.copysign(self._MAX_SPEED, body_vel)
 
-            body_curv = self.cmd[1]
-            if abs(body_curv) > self._MAX_CURVATURE:
-                body_curv = math.copysign(self._MAX_CURVATURE, body_curv) 
-            
+            body_curve = self.cmd[1]
```
hmccarty, line 98: This is a misleading variable name. Previous variable name stands for curvature, indicating the inverse of a turning radius.

rktl_sim/nodes/simulator
```diff
@@ -245,33 +218,71 @@ class Simulator(object):
...
-        
+
+        self.reset_cb()
         self.reset_lock.release()
+
         return DeleteCarResponse(res)
 
-    def reset_cb(self, _):
-        """Resets simulator"""
+    def reset_cb(self, reset_sim=True):
```
hmccarty, line 231: Can't think of a scenario where this function doesn't reset the sim. We should maybe update the docstring as well. Also this is a ROS service callback, which requires we have an additional argument for the request. It's an empty request, hence the underscore, but it's still required.

rktl_sim/src/simulator/car.py
```diff
@@ -104,23 +124,54 @@ def get_pose(self, noise=None):
                 pos = np.random.normal(pos, noise['pos'])
                 orient = np.random.normal(orient, noise['orient'])
 
-        return (pos, p.getQuaternionFromEuler(orient))
+        return pos, p.getQuaternionFromEuler(orient)
 
     def get_velocity(self):
+        """
+        @return: the velocity of the car
+        """
```
hmccarty, line 132: We can probably make this more informative. Right now it doesn't provide more information than the function name. Maybe detail on the units / framing.

rktl_sim/src/simulator/sim.py
```diff
@@ -74,20 +89,37 @@ def __init__(self, props, urdf_paths, spawn_bounds, render_enabled):
 
         if 'engine' in self.props and self.props['engine'] is not None:
             p.setPhysicsEngineParameter(**self.props['engine'])
+        # TODO: why gravity not accurate?
```
hmccarty, line 92: See previous comment about TODOs. This probably should be parameterized, it's been solid up to this point, but we want to likely include that in domain randomization.

rktl_sim/src/simulator/sim.py
```diff
@@ -74,20 +89,37 @@ def __init__(self, props, urdf_paths, spawn_bounds, render_enabled):
...
         p.setGravity(0, 0, -10)
 
     def configure_dynamics(self, body_id, body_type):
+        """
+        Configure the dynamics of the car.
```
hmccarty, line 97: Need to be really careful about comments that restate function or variable names. Also strive to make comments concise, yet elucidating.

hmccarty (changes requested):

hmccarty:
Getting closer. Key idea is that less is often more when it comes to large code bases. We should be conservative with our changes to avoid unnecessary complexity.

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -40,13 +42,13 @@ class ROSInterface(Env):
...
-            launch_file: if training, launch file to be used (ex: ['rktl_autonomy', 'rocket_league_train.launch'])
-            launch_args: if training, arguments to be passed to roslaunch (ex: ['render:=true', rate:=10])
-            run_id: if training, used to prevent deadlocks. if logging, run_id describes where to save files. Default is randomly generated
+        """
+        initializes the rospy interface
+        @param node_name: desired name of this node in the ROS network
+        @param eval: set true if evaluating an agent in an existing ROS env, set false if training an agent
+        @param launch_file: if training, launch file to be used (ex: ['rktl_autonomy', 'rocket_league_train.launch'])
+        @param launch_args: if training, arguments to be passed to roslaunch (ex: ['render:=true', rate:=10])
+        @param run_id: if training, used to prevent deadlocks. if logging, run_id describes where to save files.
```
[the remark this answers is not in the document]
Abuynits: To clarify, should I keep these in or remove them and create docstring comments? (my ide autogenerates them and I can fill in all the specifications for each variable and return statement [if one exists]).

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -15,16 +15,18 @@
...
     pass
 
+
 class ROSInterface(Env):
     """Extension of the Gym environment class for all specific interfaces
     to extend. This class handles logic regarding timesteps in ROS, and
     allows users to treat any ROS system as a Gym environment once the
     interface is created.
-
+    # IMPORTANT: All the below abstract methods marked with @abstractmethod must be implemented by subclasses
```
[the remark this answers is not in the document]
Abuynits: Yes, that makes sense.

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
```diff
@@ -76,10 +78,11 @@ def __init__(self, node_name='gym_interface', eval=False, launch_file=None, laun
             ros_id = roslaunch.rlutil.get_or_generate_uuid(None, False)
             roslaunch.configure_logging(ros_id)
             launch_file = roslaunch.rlutil.resolve_launch_arguments(launch_file)[0]
-            launch_args = [f'render:={port==11311}', f'plot_log:={port==11311}'] + launch_args + [f'agent_name:={node_name}']
+            launch_args = [f'render:={port == 11311}', f'plot_log:={port == 11311}'] + launch_args + [
+                f'agent_name:={node_name}']
```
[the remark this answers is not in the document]
Abuynits: Done!

rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
```diff
@@ -232,6 +259,7 @@ def _publish_action(self, action):
 
     def _car_odom_cb(self, odom_msg):
         """Callback for odometry of car."""
+        # Get the x,y position and yaw from euler from quaternion calculation.
```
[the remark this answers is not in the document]
Abuynits: I might have forgotten to look over some files using ideas from previous comments. This completely makes sense.

rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
```diff
@@ -96,40 +111,45 @@ def __init__(self, eval=False, launch_file=['rktl_autonomy', 'rocket_league_trai
...
+        # Block simulation until environment is ready.
         if not eval:
             rospy.wait_for_service('sim_reset')
 
     @property
     def action_space(self):
-        """The Space object corresponding to valid actions."""
+        """
+        The Space object corresponding to valid actions.
+        """
```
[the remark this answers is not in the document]
Abuynits: Makes sense.

rktl_autonomy/src/rktl_autonomy/rocket_league_interface.py
```diff
@@ -96,40 +111,45 @@ def __init__(self, eval=False, launch_file=['rktl_autonomy', 'rocket_league_trai
         rospy.Subscriber('ball/odom', Odometry, self._ball_odom_cb)
         rospy.Subscriber('match_status', MatchStatus, self._score_cb)
 
-        # block until environment is ready
+        # Block simulation until environment is ready.
```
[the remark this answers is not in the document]
Abuynits: cool, thanks for the explanation!

rktl_sim/nodes/simulator
```diff
@@ -15,130 +16,89 @@ from std_srvs.srv import Empty, EmptyResponse
...
+    """ROS wrapper for the simulator
+    Handles the simulator setup: sim mode, sim rate, sensor noise, urdf paths, rendering, etc.
+    Cars and ball objects and their functionality are handed off to the sim.py.
+
+    Two types of parameters: configuration-based and instance-based.
+    Configuration-based parameters determine the kinematics and dynamics of the sim.
+        A few examples: car size, sim timestep, friction, etc.
+    Instance-based parameters determine the initial state of a new run.
+        A few examples: car pose, ball pose, etc.
+    On startup, we load both configuration and instance parameters and create an initial setup.
```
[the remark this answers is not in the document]
Abuynits: W

rktl_sim/src/simulator/car.py
Abuynits: I renamed it to body_curvature. Does this work, or should I keep it at body_curv?

rktl_planner/nodes/path_planner
```diff
@@ -18,18 +18,22 @@ def create_simple_path_req(car_odom, ball_odom, goal_pos):
     req.bezier_segment_duration.data = rospy.Duration(0.5)
     req.linear_segment_duration.data = rospy.Duration(0.01)
 
-    # Target 0 (car pos)
+    # Target 0 (car pos)3
+    #pose: position of the car
+    # the
```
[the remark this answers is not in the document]
Abuynits: I think I messed around with path_planner and I forgot to look over the file again and add correct documentation. Can you take another look over this file?

rktl_planner/nodes/path_planner
```diff
@@ -18,18 +18,22 @@ def create_simple_path_req(car_odom, ball_odom, goal_pos):
...
     pose0.position.z = 0.0
     pose0.orientation = car_odom.pose.pose.orientation
+    #append the new pose to the list of positions
     req.target_poses.append(pose0)
     req.target_durations.append(Duration(data=rospy.Duration(5.0)))
 
     # Target 1 (ball pos)
     final_vec_x = goal_pos[0] - ball_odom.pose.pose.position.x
     final_vec_y = goal_pos[1] - ball_odom.pose.pose.position.y
+    #^save the velocities to prevent overlap in them when multithread processes
```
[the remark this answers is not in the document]
Abuynits: I just added a personal note a long time ago and I forgot to remove it. I deleted it.

rktl_planner/nodes/path_planner
```diff
@@ -49,6 +53,7 @@ def create_simple_path_req(car_odom, ball_odom, goal_pos):
     pose2.position.x = pose1.position.x + final_vec_x / final_vec_len / 2
     pose2.position.y = pose1.position.y + final_vec_y / final_vec_len / 2
     pose2.position.z = 0.0
+    #TODO: what is the point of this line
```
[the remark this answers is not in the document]
Abuynits: I removed this todo. Should I look into removing them from global commit history via bfg Repo-Cleaner, if this is something you think can help?

rktl_sim/nodes/simulator
```diff
@@ -15,130 +16,89 @@ from std_srvs.srv import Empty, EmptyResponse
...
+# Local libraries.
 import simulator
 from rktl_sim.srv import CreateCar, CreateCarResponse, DeleteCar, DeleteCarResponse
 from rktl_msgs.msg import MatchStatus, ControlCommand, ControlEffort
 
+
 class SimulatorMode(Enum):
+    # Choose between two types of models:
+    # Ideal (no sensor noise) or realistic (simulate sensor noise).
     IDEAL = 1
```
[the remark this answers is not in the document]
Abuynits: Ye, this makes a lot of sense. Looking in to the car.py class, I see that it also introduces variance in the pos and orient of the car and ball (via the sensor noise). I also saw that in the simulator class, it controls certain publishers where the ideal mode would publish the pose and orient of the ball and car early, while the realistic mode would publish it on time ( and therefore have a delay of the true position of the cars and ball at that time). Is this correct? i tried to summarize it in my comment.

rktl_sim/nodes/simulator
```diff
@@ -15,130 +16,89 @@ from std_srvs.srv import Empty, EmptyResponse
...
     REALISTIC = 2
 
+
 class Simulator(object):
-    """ROS wrapper for the simulator."""
+    """ROS wrapper for the simulator
+    Handles the simulator setup: sim mode, sim rate, sensor noise, urdf paths, rendering, etc.
+    Cars and ball objects and their functionality are handed off to the sim.py.
```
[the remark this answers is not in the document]
Abuynits: Cool, I listed down the default parameters, and added more notes on the configuration vs instance based parameters!

rktl_sim/config/simulation.yaml
```diff
@@ -55,12 +55,12 @@ dynamics:
 
 sensor_noise:
   car:
-    pos: [0.01, 0.01, 0.00]
-    orient: [0.0, 0.0, 0.09]
+    pos: [ 0.01, 0.01, 0.00 ]
+    orient: [ 0.0, 0.0, 0.09 ]
```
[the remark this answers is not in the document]
Abuynits: Cool. my auto formater adds them in for some reason. I undid them!

rktl_sim/src/simulator/car.py
Abuynits: I totally agree that this should be more informative. I looked over the car's get_velocity function and it returns linear and angular velocity. Out of curiosity, what do the link_state[6:8] commands do, as I feel it would be a good thing to jot down.

rktl_autonomy/nodes/rocket_league_agent
```diff
@@ -11,18 +11,20 @@ from stable_baselines3 import PPO
 from os.path import expanduser
 import rospy
 
-# create interface (and init ROS)
+# Create interface (and init ROS).
```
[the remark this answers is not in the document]
Abuynits: Ok, I think I went over all of my files and I changed the comments that are not in a docstring to not have punctuation.

rktl_sim/nodes/simulator
Abuynits: Ok, I think I know what needs to be done, but before I do it, I want to check that this is the right way to go. Instead of traversing from simulator to sim to car/ball objects, I should use rospy.get_param in the car/ball objects to constantly refer to parameters in the rospy. That way when I call the reset_cb function in the simulator, using the rospy.set_param, I will change the parameters in ropsy, and hence when the car and ball objects grab from ropsy via rospy.get_param, their own parameters will get reset.

I found an example of this code where something like this happens:
`car_param = [
  {
    "name": "car0",
    "init_pose": {
      "pos": [
        init_car_odom.pose.pose.position.x,
        init_car_odom.pose.pose.position.y,
        0.06,
      ],
      "orient": [
        0.0, 0.0, car_euler[2]
      ]
    }
  }
]

rospy.set_param(f'{sim_name}/cars', car_param)`

This will be called in the reset_cb function instead of passing parameters to sim.py, car.py and the ball object and physically changing the old parameters in each object. Is this correct?

Because if it is, it will save a bunch of time as by using rospy to link the simulator to the sim /car and ball objects, it will clean up a lot of code that currently does the same functionality. It will also make it easier to follow what happens in the sim.py and the car/ball objects as there will not be a large trail of variables heading from the simulator to the sim to the ball objects. It will also be very clean.

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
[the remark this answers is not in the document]
hmccarty: Def keep, these are good.

rktl_sim/src/simulator/car.py
hmccarty: Either of those options should be fine

rktl_sim/nodes/simulator
hmccarty: We resolved this over discord, let me know if any questions remain

rktl_sim/src/simulator/car.py
Abuynits: W

rktl_autonomy/src/rktl_autonomy/_ros_interface.py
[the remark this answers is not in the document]
Abuynits: cool!

PR #150, opened 2022-09-16
Time step config

jcrm1:
Adds a config option to use p.setTimeStep (off by default)  
I found this while looking for a way to slow down the simulation and watch movement more closely. However, since I don't have a very deep understanding of pybullet yet, I don't know if it messes with simulation outside of speed, so I left it off by default.  
0.001 is a good starting point if anyone is looking to test it.

### DIFF
--- a/rktl_sim/nodes/simulator
+++ b/rktl_sim/nodes/simulator
@@ -81,7 +81,7 @@ class SimulatorROS(object):
             self.check_urdf(path)
 
         # Creating physics simulator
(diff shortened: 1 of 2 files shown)


jcrm1:
Just looked at the failed check- log just says "Failed" but also that all packages built. PR builds & runs sim correctly on my local machine. Does anyone have an idea of what the issue is?

jcrm1:
Stale; closing

PR #153, opened 2022-09-30
Add YAML Editor and update .gitignore

jcrm1:
Adds a visualizer/editor for YAML 2D arrays. This is a draft because we have since changed to two 1D arrays, and I am working on changing it to support that instead. This is using modular_training as the base because that's what it's relevant for.  
Note: If we ever decide we want to start editing a bunch more parameters, it will be significantly more work to add than if all values were stored in a 2D array ([[win1, lose1],[win2, lose2]]: [[1, 2], [3, 4]]) because of how many UI elements will need to be added (text fields for each array, code to copy data from fields to table and to output)


dgerblick:
Be sure to add a readme (either as part of the exisiting `rktl_autonomy` one or in the `YAMLEditor` subfolder before making the final change. Also, is it possible to change the folder layout for this? I haven't used IntelliJ in years so I don't know how to do that, but it's going to be a pain to navigate 6 nested empty directories.

Last thing. For future reference, you can just create a new branch for changes, you don't need to fork the entire repo. Either one is fine really, so do whichever way is easier for you.

jcrm1:
README.md added, and for the layout thing, your IDE will clean it up and sort by package- the folder tree thing is standard. Last, is there a package/bundle ID for ARC? If nothing else, I can just use `com.purduearc`. This should be the last change required before it's ready.

jcrm1:
Using com.purduearc for now

jcrm1:
Backend is updated. Merging editor

PR #155, opened 2022-10-14
Accept more rewards/penalties as arrays

jcrm1:
Also sets defaults as arrays. Needs testing


jcrm1:
Added in other branch

PR #156, opened 2022-10-16
Partial Sim Refactoring

Abuynits:


jcrm1:
Works with modular training

PR #168, opened 2023-03-25
Abuynits/jcrm1 sim refactoring

dgerblick:
What is the current state of this branch? What does it change and what can we do to ensure it doesn't break anything if merged?


jcrm1:
Error in check is
```
could not connect to display
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.
```

dgerblick:
That sounds like a bug in the test case, it shouldn't be trying to open a GUI. Which package is it in?

jcrm1:
It's from the only test in `rktl_autonomy`. Part of the test passes `launch_args=['render:=false']` to a `RocketLeagueInterface` constructor, which passes those to the `ROSInterface` constructor, which passes `launch_args` to the `rocket_league_train.launch` launch file, which passes the `render` arg to `rocket_league_sim.launch`, which defaults `render` to true and uses it as follows:
```
<include if="$(arg render)" file="$(find rktl_sim)/launch/visualizer.launch"/>
```
and
```
<node pkg="rqt_gui" type="rqt_gui" name="rqt_gui" args="--perspective-file $(find rktl_launch)/rqt/rktl.perspective"/>
```
Note that the second use is not wrapped with an `if`. Added the if & testing in this branch.

jcrm1:
Errors are a timeout in rktl_autonomy and a (silent?) failure in rktl_sim. Will give it a look when I get a chance.

jcrm1 (approved):

